PR #2, opened 2018-03-28
Timezone offset work

jbradl11:
- Add an optional parameter for "execution DateTime". If a DateTime is not passed in, then the Context will generate one on creation. DateTimes in the execution engine will use the timezone offset provided by the "execution DateTime" if one is not provided for any particular DateTime.
- The CQL functions `Now` and `Today` use the "execution DateTime" as their source.
- Updated CalculateAge tests to use a static Date to make testing more reliable.
- Re-generated the cql4browsers

**Submitter:**
- [x] This pull request describes why these changes were made.
- [x] This PR is into the correct branch.
- [x] JIRA ticket for this PR: https://jira.mitre.org/browse/BONNIE-1300
- [ ] JIRA ticket links to this PR: 
- [x] Code diff has been done and been reviewed (it **does not** contain: additional white space, not applicable code changes, debug statements, etc.)
- [x] Tests have been run locally and pass (remember to update Gemfile when applicable)
- [x] Automated regression test(s) pass

If JIRA tests were used to supplement or replace automated tests:
- [x] JIRA test links: N/A
- [x] Justification for using JIRA tests: N/A
- [x] JIRA tests have been added to sprint: N/A

**Reviewer 1:**

Name:

If JIRA tests were used to supplement or replace automated tests:
- [ ] JIRA tests have been run and pass
- [ ] You agree with the justification for use of JIRA tests or have provided input on why you disagree

**Reviewer 2: Luke**

Name:

If JIRA tests were used to supplement or replace automated tests:
- [ ] JIRA tests have been run and pass
- [ ] You agree with the justification for use of JIRA tests or have provided input on why you disagree


src/runtime/context.coffee
```diff
@@ -193,26 +210,24 @@ module.exports.Context = class Context
...
   findRecords: ( profile) ->
     @patient?.findRecords(profile)
 
-
-
 module.exports.PopulationContext = class PopulationContext extends Context
 
-  constructor: (@library, @results, codeService, parameters) ->
-    super(@library,codeService,parameters)
+  constructor: (@library, @results, codeService, parameters, @executionDateTime = dt.DateTime.fromDate(new Date())) ->
```
losborne: The default value for this constructor is not covered by the tests:
image

src/datatypes/datetime.coffee
```diff
@@ -23,6 +23,7 @@ module.exports.DateTime = class DateTime
       null
 
   @fromDate: (date, timezoneOffset) ->
+    if (date instanceof DateTime) then return date
```
losborne: On line 89 below there is the function:
```
convertToTimezoneOffset: (timezoneOffset = 0) ->
  d = DateTime.fromDate(@toJSDate(), timezoneOffset)
  d.reducedPrecision(@getPrecision())
```

With this new logic, that will no longer alter the timezoneOffset of a provided `DateTime`.  Should this instead first set the timezoneOffset before returning `date`?

src/elm/clinical.coffee
```diff
@@ -116,7 +116,7 @@ module.exports.CalculateAge = class CalculateAge extends Expression
 
   exec: (ctx) ->
     date1 = @execArgs(ctx)
-    date2 = dt.DateTime.fromDate(new Date())
+    date2 = dt.DateTime.fromDate(ctx.getExecutionDateTime())
```
losborne: Right now, `ctx.getExecutionDateTime()` can return `null`, in which case I believe `fromDate` would enter its `else` block and throw an error.

src/runtime/context.coffee
```diff
@@ -62,6 +63,22 @@ module.exports.Context = class Context
     else if @parent?
       @parent.getParentParameter name
 
+  getTimezoneOffset: ->
+    if @executionDateTime?
+      @executionDateTime.timezoneOffset
+    else if @parent?.getTimezoneOffset?
```
losborne: The else if branch is not covered by the tests:
image

src/elm/clinical.coffee
jbradl11: So, it was clear from the beginning that we needed to make sure there is always an `executionDateTime` even if its the default one. As the engine is currently built, there will always be a valid 'executionDateTime' if `PatientContext` or `PopulationContext` are used, because all other contexts branch from the base `Patient` and `Population` contexts. So if there is a case where `ctx.getExecutionDateTime()` returns null, then errors should occur. Where and what those errors should be / look like we can discuss in the morning.

src/datatypes/datetime.coffee
jbradl11: I believe `@toJSDate()` first converts the DateTime to a javascript `Date` object before it gets passed into the `fromDate` function. So the new logic (I believe) will not alter the functionality of `convertToTimezoneOffset`.

src/runtime/context.coffee
```diff
@@ -14,7 +15,7 @@ module.exports.Context = class Context
     # TODO: If there is an issue with number of parameters look into cql4browsers fix: 387ea77538182833283af65e6341e7a05192304c
     @checkParameters(_parameters) # not crazy about possibly throwing an error in a constructor, but...
     @_parameters = _parameters
-
+
```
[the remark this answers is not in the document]
jbradl11: Good catch.

src/runtime/context.coffee
jbradl11: So I don't think we have any tests for `Population Context` and I am not sure if CQL measures will behave correctly if and when people try to use `Population Context`. I only added this to it because I felt it was a required field, and its an exact copy of `Patient Context`. I think adding support / tests for Population Context should be a future task.

src/elm/clinical.coffee
jbradl11: @cmoesel any thoughts on this?

src/runtime/context.coffee
cmoesel: I agree -- I wouldn't hold up the PR for this.

src/elm/clinical.coffee
cmoesel: Perhaps the `getExecutionDateTime` function should throw an error if one is not set?  Does that sound reasonable?

src/runtime/context.coffee
cmoesel: I think @jbradl11 added a test to cover this after @losborne made this comment.

src/example/browser/simple-browser-support.coffee
```diff
@@ -1,6 +1,6 @@
 window.cql = require '../../cql'
 
-window.executeSimpleELM = (elm, patientSource, valueSets, libraryName, version, parameters = {}) ->
+window.executeSimpleELM = (elm, patientSource, valueSets, libraryName, version, executionDateTime, parameters = {}) ->
```
cmoesel: So... in this same directory, there is an example HTML file that invokes `executeSimpleELM`.  It seems it's probably been broken for a while (not your fault), but can you update it to invoke `executeSimpleELM` with the right function signature?  It would be good to have an example of how the `executionDateTime` should be passed in.

https://github.com/cqframework/cql-execution/blob/ea6806eccf059722ded252d1a66a4ef535280f3d/src/example/browser/cql4browsers.html#L13

src/runtime/executor.coffee
```diff
@@ -23,17 +23,17 @@ module.exports.Executor = class Executor
       patientSource.nextPatient()
     r
 
-  exec: (patientSource) ->
-    Results r = @exec_patient_context(patientSource)
+  exec: (patientSource, executionDateTime) ->
```
cmoesel: There are a couple of examples in the source code that invoke this function.  Can you update them to pass in the `executionDateTime`?  I think it would be good to show this:

https://github.com/cqframework/cql-execution/blob/ea6806eccf059722ded252d1a66a4ef535280f3d/src/example/exec-age.coffee#L46

https://github.com/cqframework/cql-execution/blob/ea6806eccf059722ded252d1a66a4ef535280f3d/src/example/exec-cms146v2_CQM.coffee#L102

src/elm/clinical.coffee
jbradl11: That works for me. I will make this change now.

src/example/browser/simple-browser-support.coffee
jbradl11: I am thinking it might be better for Bonnie to create a new entry point into the engine (somewhere not under `example`) And then we change this file back to what it was before. Thoughts? I would say this would also be a different PR.

cmoesel: I think it probably would be good for Bonnie dependencies not to be in an _example_ directory.  That said, I think it would also be good for the example to show how to do this. ;-)

But if it's trickier than it looks, just go ahead and leave it like this for this PR (since I think it's been broken for a while).

jbradl11: Completely agree! I'll take a crack at it and see if I can get it working without too much effort.

test/elm/clinical/test.coffee
```diff
@@ -158,17 +160,20 @@ describe 'CalculateAge', ->
...
 
     # this is getting the possible number of months in years with the addtion of an offset
     # to get the correct number of months
-    month_offset = if @today.getMonth() == 5 && @today.getDate() < 17 then 6 else 5
-    @full_months = ((@today.getFullYear() - 1980) * 12) + (@today.getMonth() - month_offset)
+    month_offset = if @today.month == 5 && @today.getDate() < 17 then 6 else 5
+    @full_months = ((@today.year - 1980) * 12) + (@today.month - month_offset)
 
-    @timediff = @today - @bday # diff in milliseconds
+    @timediff = @today.toJSDate() - @bday # diff in milliseconds
```
cmoesel: Since we're now essentially hard-coding a date as `today`, can't we finally get rid of a lot of the hokey stuff above?  Can't we replace much of it with constants?

test/elm/datetime/test.coffee
```diff
@@ -154,33 +156,78 @@ describe 'Time', ->
 describe 'Today', ->
   @beforeEach ->
     setup @, data
+    @ctx = new PatientContext(@ctx.library, @ctx.patient, @ctx.codeService, @ctx.parameters)
```
cmoesel: Wouldn't it be better to pass in an actual hard-coded execution date time (e.g. `2012-04-15T03:22:56.123-04:00` and then below test against the expected constants (e.g., `2012` rather than `@ctx.getExecutionDateTime().year`)?

cmoesel: Also same for other similar tests you changed?

test/elm/datetime/test.coffee
```diff
@@ -154,33 +156,78 @@ describe 'Time', ->
...
+    today.day.should.equal @ctx.getExecutionDateTime().day
+    today.timezoneOffset.should.equal "0"
     should.not.exist(today[field]) for field in [ 'hour', 'minute', 'second', 'millisecond' ]
 
+  it 'should throw an exception because no execution datetime has been set', ->
+    try
+      @ctx = new PatientContext(@ctx.library, @ctx.patient, @ctx.codeService, @ctx.parameters, DT.DateTime.fromDate(new Date(), '0'))
+      @ctx.executionDateTime = @ctx.executionDateTime = null
+      @todayVar.exec(@ctx).should.equal "No Execution DateTime has been set"
+    catch
```
cmoesel: Unless I'm misunderstanding it, I don't think this test actually does what it says it does.

test/elm/datetime/test.coffee
```diff
@@ -154,33 +156,78 @@ describe 'Time', ->
...
+
 describe 'Now', ->
   @beforeEach ->
     setup @, data
+    @ctx = new PatientContext(@ctx.library, @ctx.patient, @ctx.codeService, @ctx.parameters)
```
cmoesel: Again, same as above, this is a great opportunity to pass in a constant datetime and test against it.

test/elm/datetime/test.coffee
cmoesel: On second read, it _does_ test it, but it's a little confusing as written.  If it didn't throw, it would fail because `@todayVar` can never equal "No Execution DateTime has been set" -- but since it does throw it goes to the empty catch.

At the least, I'd prefer to see a comment like:
```
// the following line should never be executed because the line above is expected to throw
```

But ideally, line 184 would be something more like:
```
should(() => @todayVar.exec(@ctx)).throw(/No Execution DateTime has been set/)
```

PR #3, opened 2018-03-30
Return null in indexer instead of out of bounds exception

mayerm94:
https://jira.mitre.org/browse/BONNIE-1345
From section 9.10.12 of the cql spec
The indexer ([]) operator returns the element at the indexth position in a list.
Indexes in lists are defined to be 0-based.
If the index is less than 0, or greater than the number of elements in the list, the result is null.
If either argument is null, the result is null.


test/elm/string/test.coffee
```diff
@@ -122,11 +121,11 @@ describe 'Indexer', ->
   it 'should get letter at index', ->
     @helloWorldSix.exec(@ctx).should.equal 'o'
 
-  it 'should not error on index 0 (no longer out of bounds)', ->
-    should(() => @helloWorldZero.exec(@ctx)).not.throw(ArrayIndexOutOfBoundsException)
+  it 'should not return null on index 0', ->
```
jbradl11: How come `(no longer out of bounds)` was removed?

PR #9, opened 2018-04-17
Equivalent fix

losborne:
This PR fixes an issue where `equivalent` was not properly comparing objects.  It also refactors both `equals` and `equivalent` to reduce duplication between them. 

An additional issue with `equivalent` was uncovered, in which `undefined` as an input was not behaving appropriately, so this PR also fixes that.

**Submitter:**

Branch | Coverage
-- | -- 
`master` | 80.91% Statements 4476/5532 <br> 72.82% Branches 887/1218 <br> 25.18% Functions 754/2994 <br> 80.43% Lines 4296/5341 
`equivalent_fix` | 80.95% Statements 4478/5532  <br> 72.91% Branches 888/1218  <br> 25.18% Functions 754/2994 <br> 80.47% Lines 4298/5341 

**Reviewer:**

Name: @cmoesel 

Name: @jbradl11 


src/util/comparison.coffee
```diff
@@ -43,11 +43,48 @@ module.exports.greaterThanOrEquals = (a, b, precision = DateTime.Unit.MILLISECON
     else null
 
 module.exports.equivalent = equivalent = (a, b) ->
-  if a == null && b == null
-    return true
-  return a.hasMatch b if typeof a.hasMatch is 'function'
+  return true if not a? and not b?
+  return false unless a?
+  return false unless b?
```
jbradl11: I think you can change this to: `return false unless a? and b?`

src/util/comparison.coffee
```diff
@@ -43,11 +43,47 @@ module.exports.greaterThanOrEquals = (a, b, precision = DateTime.Unit.MILLISECON
...
+  return codesAreEquivalent(a, b) if isCode(a)
+
+  [aClass, bClass] = getClassOfObjects(a, b)
+  switch aClass
+    when '[object Array]'
+      return compareEveryItemInArrays(a, b, equivalent)
+    when '[object Object]'
+      return compareObjects(a, b, equivalent)
+
   return equals a, b
```
jbradl11: I see that we are still falling back to `equals` as a last resort. I just want to make sure that those are **known** specialized cases. Since in the Spec it says `Note  that this operator will always return true or false, even if either or both of its arguments are null, or contain null components`. And the `equals` operator can return null in some cases. Specifically around date/time comparisons.

losborne: `equals` can return null in two cases. The first case is
```coffeescript
return null unless a? and b?
```
This case will never happen because equivalent returns false before it falls through to `equals`:
```coffeescript
 return false unless a? and b?
```

The second case is 
```coffeescript
[aClass, bClass] = getClassOfObjects(a, b)
return false if aClass isnt bClass

switch aClass
...
when '[object Array]'
  return null if a.includes(null) or a.includes(undefined) or b.includes(null) or b.includes(undefined)
```
This case will also never fall through to `equals` because the same code appears in `equivalent` before the call to `equals`:

```coffeescript
[aClass, bClass] = getClassOfObjects(a, b)
switch aClass
  when '[object Array]'
    return compareEveryItemInArrays(a, b, equivalent)
```

Date/Time Comparisons, when they fall through to `equals`, can currently never return `null`.  If a future fix will end up with `equals` returning `null` in the case of `'[object Date]`, then another `when` statement in `equivalent` needs to be added to catch `[object Date]` before it falls through to equal.

losborne: Writing that response made me realize that the line
```coffeescript
return false if aClass isnt bClass
```
isn't present in `equivalent`. 

I'll try to add a failing unit test then add that line.

losborne: Added tests, which increased coverage, but wasn't able to get anything to break.

src/util/comparison.coffee
```diff
@@ -43,11 +43,49 @@ module.exports.greaterThanOrEquals = (a, b, precision = DateTime.Unit.MILLISECON
...
-  if a == null && b == null
-    return true
-  return a.hasMatch b if typeof a.hasMatch is 'function'
+  return true if not a? and not b?
+  return false unless a? and b?
+
+  return codesAreEquivalent(a, b) if isCode(a)
+
+  [aClass, bClass] = getClassOfObjects(a, b)
+  # return false if aClass isnt bClass
```
jbradl11: Is this a comment or commented out code?

losborne:
@cmoesel The Bonnie regression test looks good for this PR's `cql4browsers.js`.  The only changes are positive and expected.  This should be good to merge.

PR #12, opened 2018-04-23
Change union to not remove all null values

mayerm94:
https://jira.mitre.org/browse/BONNIE-1371
Do distinct will now not discard null values.

**Submitter:**
- [x] Regression test passes

Expected Regression Between Bonnie 2.1 and This in cql_qdm_patientapi
hmdmlantana
====Patients Passing Regression===========
+: Population: Stratification 3  SECOND BRANCH HAS MORE PASSES (24 < 33)
Measures with no differences: 14/15
====Code Coverage Percentage Regression===========
+: CMS ID: CMS111v7-0 SECOND BRANCH HAS MORE COVERAGE (98 < 100)
+: CMS ID: CMS55v7-0 SECOND BRANCH HAS MORE COVERAGE (98 < 100)
+: CMS ID: CMS32v8-0-1-2 SECOND BRANCH HAS MORE COVERAGE (82 < 100)
Measures with no differences: 9/12

**Reviewer:**

Name:


test/elm/list/test.coffee
```diff
@@ -87,6 +87,9 @@ describe 'Union', ->
 
   it 'should remove duplicate elements (according to CQL 1.2 spec)', ->
     @oneToFiveOverlapped.exec(@ctx).should.eql [1, 2, 3, 4, 5]
+
```
losborne: There is some whitespace at the beginning of this line that should be deleted.

test/elm/list/test.coffee
```diff
@@ -384,6 +387,9 @@ describe 'Distinct', ->
 
   it 'should do nothing to an array of distinct tuples', ->
     @noDupsTuples.exec(@ctx).should.eql [{ hello: 'world' }, { hello: 'cleveland' }]
+
```
losborne: There is some whitespace at the beginning of this line that should be deleted.

src/util/comparison.coffee
```diff
@@ -45,6 +45,8 @@ module.exports.greaterThanOrEquals = (a, b, precision = DateTime.Unit.MILLISECON
 module.exports.equivalent = equivalent = (a, b) ->
   if a == null && b == null
     return true
+  else if a == null || b == null
```
losborne: `a == null` will get transpiled to `a === null` in JavaScript, which will fail for `undefined` input.  What we really want is `a == null` in Javascript which is `a?` in CoffeeScript.

Lines 46-49 on this branch should go away when the merge conflict with https://github.com/cqframework/cql-execution/pull/9 gets resolved.  It might be best to replace them here just to reduce the risk that these lines make it through instead somehow.

They should instead be:
```
return true if not a? and not b?
return false unless a? and b?
```

PR #22, opened 2018-05-09
Cql 1.2 alignment 2 master

losborne:
This PR adds some fixes for section 9.3 and 9.5 of the CQL spec (1 commit for each, see their extended commit messages for squashed commit messages).

Related JIRA Links: 
- Section 9.3 https://jira.mitre.org/browse/BONNIE-1380
- Section 9.5 https://jira.mitre.org/browse/BONNIE-1385 

**Submitter:**

Branch | Coverage
-- | --
master | 81.53% Statements 4551/5582<br>75.52% Branches 947/1254<br>25.98% Functions 785/3021<br>81.08% Lines 4371/5391
cql_1.2_aligntment_2_master |81.71% Statements 4605/5636<br>75.69% Branches 962/1271<br>26.24% Functions 795/3030<br>81.26% Lines 4422/5442

- [x] All dependent libraries are appropriately updated or have a corresponding PR related to this change: https://jira.mitre.org/browse/BONNIE-1493
- [x] Bonnie regression test passes

**Reviewer:**

Name: @cmoesel 

**Reviewer:**

Name: @hossenlopp 


test/elm/comparison/test.coffee
```diff
@@ -108,13 +108,13 @@ describe 'NotEqual', ->
     @aLtB_Quantity_diff.exec(@ctx).should.be.true()
 
   it 'should be null for 5 Cel != 4 m', ->
-    should(@aGtB_Quantity_incompatible.exec(@ctx)).be.null()
+    should(() => @aGtB_Quantity_incompatible.exec(@ctx)).throw("Incompatible Types 'm' and 'Cel'")
```
hossenlopp: The title of this test does not match what the test is doing. This is an issue with all remaining test changes in this file.

src/elm/comparison.coffee
```diff
@@ -16,9 +16,9 @@ module.exports.Less = class Less extends Expression
       args[0].lessThan args[1]
     catch error
       if error instanceof IncompatibleTypesException
-        return null
-      else
         throw error
+      else
+        return null
```
hossenlopp: This flip of what happens doesn't make sense to me. Before it was catching the IncompatibleTypeException and returning a null, if the error was something else then it was re-throwing the error. After this change it eats and returns null for ANY error except IncompatibleTypeException. This doesn't seem correct to me.

test/elm/comparison/test.coffee
losborne: Good catch, I can update these test descriptions.

src/elm/comparison.coffee
losborne: @mayerm94 can you field this question?  I don't mind updating if it needs to be updated.

jbradl11: Based on the spec, we shouldn't be returning null if there is an incompatible type exception.  I believe @mayerm94 discussed with Moesel what should happen if a different type of error is generated.

losborne: Relevant CQL spec section:
image

It does seem weird to me as well to return `null` for other types of errors rather than throwing.

src/datatypes/datetime.coffee
```diff
@@ -391,3 +401,81 @@ module.exports.DateTime = class DateTime
...
+                'YYYY-MM-DDThh:mm:ss+hh',
+                'YYYY-MM-DDThh:mm:ss+hh:mm',
+                'YYYY-MM-DDThh:mm:ss-hh',
+                'YYYY-MM-DDThh:mm:ss-hh:mm',
+                'YYYY-MM-DDThh:mm:ss.fff',
+                'YYYY-MM-DDThh:mm:ss.fffZ',
+                'YYYY-MM-DDThh:mm:ss.fff+hh',
+                'YYYY-MM-DDThh:mm:ss.fff+hh:mm',
+                'YYYY-MM-DDThh:mm:ss.fff-hh',
+                'YYYY-MM-DDThh:mm:ss.fff-hh:mm']
```
cmoesel: I'm noticing (and you probably noticed too) that perhaps the CQL spec is not specific enough about what formats are allowed -- particularly in regard to where `Z` or `(+-)hh:mm` can go when the DateTime does not have complete precision.

That said, I'm not 100% sure that `YYYY-MM-DDZ` is valid.  Perhaps it should be `YYYY-MM-DDTZ` since it is a component of _time_?

In addition, it seems odd to support `YYYY-MM-DDT+hh` and `YYYY-MM-DDT-hh` but not `YYYY-MM-DDT+hh:mm` and `YYYY-MM-DDT-hh:mm`.

It also seems odd to support `YYYY-MM-DDThh:mmZ` but not the other representations of offsets for minutes.

Given that the spec is not very clear, then I'd rather err on the side of accepting too much (as long as we can parse it) than accepting too little.  For that reason, I'd suggest supporting `Z`, `+hh`, `+hh:mm`, `-hh`, and `-hh:mm` suffixes  at all levels of time precision.  Would you agree?

src/elm/comparison.coffee
cmoesel: I can't imagine what other types of errors would pop up -- but if you all want to throw no matter what the error then feel free.  I think we could make an argument for either way.

cmoesel: And yeah, I think I remember a discussion with @mayerm94 and I probably said returning null on other errors was fine.  I think there are other parts of CQL where unexpected things _do_ result in null.  Like I said, we could probably argue either side.

src/elm/type.coffee
```diff
@@ -56,23 +57,36 @@ module.exports.ToDecimal = class ToDecimal extends Expression
...
 module.exports.ToQuantity = class ToQuantity extends Expression
   constructor: (json) ->
     super
 
   exec: (ctx) ->
     arg = @execArgs(ctx)
-    if arg? and typeof arg != 'undefined' then parseQuantity(arg.toString()) else null
+    if arg? and typeof arg != 'undefined'
+      quantity = parseQuantity(arg.toString())
+      return null if !quantity?
```
cmoesel: Is this to ensure `null` is returned rather than `undefined`?

src/elm/comparison.coffee
edeyoung: @mayerm94 says he reviewed w moesel. that this is the expected behavior.

src/util/math.coffee
```diff
@@ -12,6 +12,28 @@ module.exports.MAX_DATE_VALUE = MAX_DATE_VALUE = DateTime.parse("9999-12-31T23:5
...
+  throw new Error("Unable to parse Integer") if isNaN(integer)
+  throw new Error("Maximum Integer value exceeded") if integer > MAX_INT_VALUE
+  throw new Error("Minimum Integer value exceeded") if integer < MIN_INT_VALUE
+  return true
+
+module.exports.isValidDecimal = isValidDecimal = (decimal) ->
+  throw new Error("Unable to parse Decimal") if isNaN(decimal)
+  throw new Error("Maximum Decimal value exceeded") if decimal > MAX_FLOAT_VALUE
+  throw new Error("Minimum Decimal value exceeded") if decimal < MIN_FLOAT_VALUE
+  throw new Error("Maximum Decimal precision") if getDecimalPrecision(decimal) > 8
```
cmoesel: This one kind of concerns me.  If we do math, we may get a decimal with greater precision than 8 digits.  I don't think that's an error condition / invalid decimal.  Perhaps we could truncate at 8 digits of precision, but I wouldn't want to throw an error.

I actually was kind of choosing to interpret the 8 digits of precision as a requirement to support at least 8 digits of precision, and that when `successor` / `predecessor` is used, we should move up or down to the next 8th digit of precision.

src/elm/comparison.coffee
cmoesel: Yeah, so to repeat -- @mayerm94 did review w/ me, and he did as we discussed.  TBH, I hadn't thought of the possibility that this might also eat an invalid UCUM unit error.  I'm totally fine with just always throwing, and in hindsight, that might be the better approach.  Sorry for leading you astray, @mayerm94.

losborne: Sounds good.  I will update these accordingly.

src/datatypes/datetime.coffee
losborne: > That said, I'm not 100% sure that YYYY-MM-DDZ is valid. Perhaps it should be YYYY-MM-DDTZ since it is a component of time?

That makes sense to me.

> I'd suggest supporting Z, +hh, +hh:mm, -hh, and -hh:mm suffixes at all levels of time precision. Would you agree?

Yes, that makes sense. I'll add support for those formats.

src/elm/type.coffee
losborne: No, I believe I was just trying to type up the CQL spec and didn't notice there could be a simplification.  I can remove this line, as parseQuantity won't ever return `undefined`

src/util/math.coffee
losborne: Good point.  I think it would be good to truncate in order to align to this clause in 9.3.9:

> Note that the decimal value returned by this operator must be limited in precision and scale to the maximum precision and scale representable for Decimal values within CQL.

cmoesel: Ha.  I guess I should have checked the spec.  Agreed.

test/elm/convert/test.coffee
```diff
@@ -159,7 +159,7 @@ describe 'ToDecimal', ->
     @negativeSign .exec(@ctx).should.equal(-1.1)
 
   it "should not return decimal that is too precise", ->
-    should(() => @tooPrecise.exec(@ctx)).throw("Maximum Decimal precision")
+    @tooPrecise.exec(@ctx).should.equal(0.44444444)
```
cmoesel: The name of this test `it "should not return decimal that is too precise"` isn't right anymore.

src/util/math.coffee
```diff
@@ -12,6 +12,33 @@ module.exports.MAX_DATE_VALUE = MAX_DATE_VALUE = DateTime.parse("9999-12-31T23:5
...
+  throw new Error("Maximum Decimal value exceeded") if decimal > MAX_FLOAT_VALUE
+  throw new Error("Minimum Decimal value exceeded") if decimal < MIN_FLOAT_VALUE
+  return true
+
+module.exports.limitDecimalPrecision = (decimal) ->
+  decimalString = decimal.toString()
+  # For decimals so large that they are represented in scientific notation, javascript has already limited
+  # the decimal to its own constraints, so we can't determine the original precision.  Leave as-is unless
+  # this becomes problematic, in which case we would need our own parseFloat.
+  if decimalString.includes('e')
```
hossenlopp: String.includes() cannot be used in IE/PantomJS this breaks the unit tests in Bonnie.

cmoesel:
Last commit looks fine.  I _still_ approve.

PR #23, opened 2018-05-09
Library Result Overwrite Fix

hossenlopp:
This resolves the issue of the PatientContext only using the results from the last PatientContext for a library. Instead of overwriting it, it merges the results gracefully from the different PatientContexts.

Additionally replaces a usage of `Array.includes` with `Array.indexOf() >= 0` to resolve compatibility issues.

**Submitter:**

Branch | Coverage
-- | --
master |81.52% Statements 4548/5579<br>75.44% Branches 943/1250<br>25.96% Functions 784/3020<br>81.07% Lines 4368/5388
bonnie_library_result_overwrite_fix | 81.53% Statements 4551/5582<br>75.52% Branches 947/1254<br>25.98% Functions 785/3021<br>81.08% Lines 4371/5391

  - https://github.com/projecttacoma/bonnie/pull/990
  - cql_qdm_patientapi PR will be made after this and #22 are merged

**CDS Reviewer:**

Name:

**Bonnie Reviewer:**

Name: @losborne

### DIFF
--- a/src/runtime/context.coffee
+++ b/src/runtime/context.coffee
@@ -138,6 +138,8 @@ module.exports.Context = class Context
     for supportLibName, supportLib of lib.library_context
       @supportLibraryLocalIds supportLib, localIdResults
 
+  # Merges the localId results for a library into the already collected results. The logic used for which result
+  # to keep is the same as the logic used above in setLocalIdWithResult, "falsey" results are always replaced.
   mergeLibraryLocalIdResults: (localIdResults, libraryId, libraryResults) ->
     for localId, localIdResult of libraryResults
       existingResult = localIdResults[libraryId][localId]
--- a/src/runtime/context.coffee
+++ b/src/runtime/context.coffee
@@ -129,10 +129,21 @@ module.exports.Context = class Context
   # Recursive function that will grab nested support library localId results
   supportLibraryLocalIds: (lib, localIdResults) ->
     # Set library identifier name as the key and the object of localIds with their results as the value
-    localIdResults[lib.library.source.library.identifier.id] = lib.localId_context
+    # if it already exists then we need to merge the results instead of overwriting
+    if localIdResults[lib.library.source.library.identifier.id]?
+      @mergeLibraryLocalIdResults localIdResults, lib.library.source.library.identifier.id, lib.localId_context
+    else
+      localIdResults[lib.library.source.library.identifier.id] = lib.localId_context
     # Iterate over any support libraries in the current support library
     for supportLibName, supportLib of lib.library_context
       @supportLibraryLocalIds supportLib, localIdResults
+
+  mergeLibraryLocalIdResults: (localIdResults, libraryId, libraryResults) ->
+    for localId, localIdResult of libraryResults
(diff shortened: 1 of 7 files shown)


src/runtime/context.coffee
```diff
@@ -129,11 +129,22 @@ module.exports.Context = class Context
...
+    else
+      localIdResults[lib.library.source.library.identifier.id] = lib.localId_context
     # Iterate over any support libraries in the current support library
     for supportLibName, supportLib of lib.library_context
       @supportLibraryLocalIds supportLib, localIdResults
 
+  mergeLibraryLocalIdResults: (localIdResults, libraryId, libraryResults) ->
+    for localId, localIdResult of libraryResults
+      existingResult = localIdResults[libraryId][localId]
+      # overwite this localid result if the existing result is "falsey". future work could track all results for each localid
```
cmoesel: I'm not sure I understand the use case.  Why are `false` and `[]` candidates to be overwritten, but other values aren't?  In the other cases (when it's set to something other than `false` or `[]`), why do we throw out the newer results, preferring the older ones?

I'm not saying this is wrong, I'd just like to understand the reasoning for this approach.

test/elm/library/test.coffee
```diff
@@ -74,3 +74,30 @@ describe 'Using CommonLib2', ->
...
+describe 'Using CommonLib and CommonLib2', ->
+  @beforeEach ->
+    setup @, data, [p1], {}, {}, new Repository(data)
+    @results = @executor.withLibrary(@lib).exec_patient_context(@patientSource)
+    @commonLocalIdObject = @results.localIdPatientResultsMap['1'].Common
+    @common2LocalIdObject = @results.localIdPatientResultsMap['1'].Common2
+
+  it "should contain TheParameter localId in the localIdMap", ->
+    theParameterLocalId = @lib.includes.common2.expressions.TheParameter.localId
+    @common2LocalIdObject[theParameterLocalId].should.exist
```
cmoesel: I think this (and others like it) needs to be:

```coffee
should.exist(@common2LocalIdObject[theParameterLocalId])
```

If `@common2LocalIdObject[theParameterLocalId]` was `null` or `undefined`, then the test as it is now would actually throw because you're trying to access `.should` on something that's `null`/`undefined`.

See: http://shouldjs.github.io/#should-exist

cmoesel: BTW... I wish I used _chai_ in this project.  Probably should switch at some point.  Oh well.

src/runtime/context.coffee
hossenlopp: Summarizing Slack conversation:

Currently we use the results for coverage in bonnie which considers a covered clause to be something that is "truthy". So we want to overwrite anything that is "falsey". This uses the same logic for overwriting that is used above in `setLocalIdWithResult`. I will update the comments to explain this better.

test/elm/library/test.coffee
hossenlopp: Good catch, those checks are never actually checked because the `()` is missing that makes them actually get called.

PR #25, opened 2018-05-30
Allow for Quantities with null values. Catch invalid values

hossenlopp:
In the Quantity constructor, allow for value to be null. Also throw errors for invalid values. ie. prevent NaN values

**Submitter:**

**Reviewer:**

Name:


test/elm/quantity/test.coffee
```diff
@@ -33,3 +33,12 @@ describe 'Quantity', ->
   it 'should allow creation of Quantity with empty string unit', ->
     should.doesNotThrow ->
       new Quantity({unit: "", value: 9})
+
+  it 'should allow for the value of the quantity to be null', ->
```
losborne: What do you think about adding another test for `undefined` like this (mostly as a way to document that this function behaves this way and to catch any changes that might work for `null` but not `undefined` in the future:

```
  it 'should convert undefined values to null', ->
    should.doesNotThrow ->
      q = new Quantity({unit: "mg", value: undefined})
      should.equal(q.value, null)
```

PR #31, opened 2018-08-01
Copy interval values to new objects in collapse function, add copy fu…

mayerm94:
…nction to interval and uncertainty. Changes made so that interval value would not be changed where it is referenced elsewhere. https://jira.mitre.org/browse/BONNIE-1628

**Submitter:**

**Reviewer:**

Name:


src/datatypes/interval.coffee
```diff
@@ -13,6 +13,16 @@ module.exports.Interval = class Interval
     isInterval:
       get: -> true
 
+  copy: ->
+    newLow = @low
+    newHigh = @high
+    if @low.copy?()
```
cmoesel: In this `if` clause (and the one below), your if condition actually _invokes_ the `copy` function -- which means that you're creating a copy (and temporarily allocating memory) just to test if the copy function exists (and if it does, then you _invoke_ copy again when you assign it).

I'd recommend you just test to see if a copy function exists like this:

```coffee
if typeof @low.copy == 'function'
```

cmoesel: Same comment applies in several places throughout this PR.

jbradl11:
Along with Moesel's comments, tests need to be added.

cmoesel:
Odd thing.  When I pulled this down and ran `yarn` (which, in turn, runs the `build-cql4browsers` task), it generated a _different_ cql4browsers.js than is checked in on this PR.

@mayerm94 - can you double-check this on your end?  Try running the cql4browsers via `yarn` and see if you get a diff.  If so, can you push it?

PR #32, opened 2018-08-02
Null test fixes

losborne:
This PR updates tests so that they assert when things should be null with `should(...).be.null()`.  Previously, `should(...).be.null` would return `undefined`, not assert (thus not fail), and the tests would silently pass.

After changing all the `null`s to `null()`s, there were a couple of failing tests, which are skipped below because they do not match the current cql spec.

(NOTE: I tested cherry picking these 2 commits into https://github.com/cqframework/cql-execution/pull/31 and tests still pass for that PR.)

**Submitter:**
- [x] All dependent libraries are appropriately updated or have a corresponding PR related to this change N/A only update to tests

**Reviewer:**

Name: @cmoesel 

**Reviewer:**

Name: @edeyoung 

### DIFF
--- a/test/elm/list/test.coffee
+++ b/test/elm/list/test.coffee
@@ -208,10 +208,10 @@ describe 'Indexer', ->
     @secondItem.exec(@ctx).should.equal 'b'
 
   it 'should NOT return null when accessing index 0', ->
-    should(() => @zeroIndex.exec(@ctx)).not.be.null()
+    should(@zeroIndex.exec(@ctx)).not.be.null()
 
   it 'should return null when accessing out of bounds index', ->
-    should(() => @outOfBounds.exec(@ctx)).be.null()
+    should(@outOfBounds.exec(@ctx)).be.null()
 
   it 'should return null if either arg is null', ->
     should(@nullList.exec(@ctx)).be.null()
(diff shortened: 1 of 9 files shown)


losborne:
@cmoesel updated and force pushed to match latest master

test/elm/list/test.coffee
```diff
@@ -289,9 +289,13 @@ describe 'Includes', ->
...
-  it 'should return null if either arg is null', ->
-    should(@nullIncluded.exec(@ctx)).be.null()
-    should(@nullIncludes.exec(@ctx)).be.null()
+  it.skip 'should return true if right arg is null', ->
+    # TODO: currently returns false
+    should(@nullIncluded.exec(@ctx)).be.true()
+
+  it.skip 'should return false if left arg is null', ->
+    # TODO: currently returns null
+    should(@nullIncludes.exec(@ctx)).be.false()
```
cmoesel: To ensure we're on the same page... you're skipping these because the current code actually produces the *wrong* answers and we don't want tests that assert them as being correct when they actually aren't?  And fixing the logic and re-enabling the tests would go into a different PR?

losborne: @cmoesel Correct. The new expected values match HL7_CQLANG_R1_STU2_2017JUL.  

List Includes:
image
List IncludedIn:
image

Even if we fixed them to match the spec, those tests are currently being translated directly to `contains` and `In`, which have different behavior than `includes` and `includedIn` after the most recent errata to the spec.

For now I think it's best to skip these tests with the expected behavior matching the spec, then it's a future task to figure out what to do about the discrepancies (if anything).

cmoesel: OK.  Sounds good to me.

cmoesel:
Do you need a second reviewer before I can merge?  It seems to me that since this only fixes test logic, and has no affect on the packaged/production code, maybe one review is enough?

losborne:
@cmoesel Yes good point--one reviewer is fine for this one.  Merge away!

PR #53, opened 2018-09-13
Bonnie v1.3 list valueset overload

losborne:
Add support for new signature for In (Valueset)
```
in(codes List<Code>, valueset ValueSetRef) Boolean
```

https://jira.mitre.org/browse/BONNIE-1719
https://github.com/HL7/cql/blob/master/spec/09-b-cqlreference.adoc#128-in-valueset

**Submitter:**
- [x] Tests have been run locally and pass ( 9 unrelated failing tests left on base branch )
(error condition is untested, but can't figure out how to get null to translate to a valueset for that condition.)
- [x] All dependent libraries are appropriately updated or have a corresponding PR related to this change N/A
- [x] regression test run and any regressions logged to jira (no regressions vs https://github.com/cqframework/cql-execution/commit/669ed2c302fc607e46222fcbdef08735781dd079)

**Reviewer:**

Name:


src/elm/clinical.coffee
```diff
@@ -27,6 +27,24 @@ module.exports.ValueSetRef = class ValueSetRef extends Expression
...
+
+  exec: (ctx) ->
+    valueset = @valueset.execute(ctx)
+    # If the value set reference cannot be resolved, a run-time error is thrown.
+    throw new Error("ValueSet must be provided to InValueSet function") unless valueset? and valueset.isValueSet
+
+    codes = @codes.exec(ctx)
+    for code in codes
+      # If the code argument is null, the result is false.
+      return false unless code?
```
daco101: I see the spec says that if the code argument is null, the result is false. But I think at that point the spec is referring to when the argument is one code, not a list. Are you sure it makes sense to return false if any code in the list is null? I think it might be better to ignore the null codes.

PR #54, opened 2018-09-18
Modified Invalueset for string overload.

jbradl11:

This PR, updates the "InValueSet" string overload to throw an error if the string "code" exists in the value set but in multiple code systems.

From the Spec:
`For the String overload, if the given valueset contains a code with an equivalent code element, the result is true. Note that for this overload, because the code being tested cannot specify code system information, if the resolved value set contains codes from multiple code systems, a run-time error is thrown because the operation is ambiguous.`

**Submitter:**
- [x] ~Code coverage has not gone down and all code touched or added is covered.~
- [x] ~All dependent libraries are appropriately updated or have a corresponding PR related to this change~

**Reviewer:**

Name:


src/datatypes/clinical.coffee
```diff
@@ -30,7 +30,22 @@ module.exports.ValueSet = class ValueSet
...
-    codesInList(toCodeList(code), @codes)
+    codesList = toCodeList(code)
+    # InValueSet String Overload
+    if codesList.length == 1 and typeof codesList[0] is 'string'
+      codes_codeSystem = {}
+      for c2 in @codes
+        if codesList[0] == c2.code
+          # Check if code exists but with different codesystem,
+          if codes_codeSystem[c2.code]? and codes_codeSystem[c2.code] != c2.system
+            throw new Error('Duplicate codes in different code systems in same valueset')
```
losborne: I suggest changing the wording of this error to use the wording from the CQL spec.  Something like:
`In (valueset) is ambiguous because resolved value set contains codes from multiple code systems.`

We could also put the code in the error message like this, which might be helpful:
`In (valueset) is ambiguous--multiple matches for #{code} found in value set with different code systems.`

PR #55, opened 2018-09-19
make toBoolean argument and string equivalence  case insensitive

mayerm94:
From Spec: 

```
The equivalent (~) operator returns true if the arguments are equivalent in value, or if they are both null; and false otherwise.
For string values, equivalence is case- and locale-insensitive.
```

toBoolean
`Note that the operator will ignore case when interpreting the string as a Boolean value.`

**Submitter:**

**Reviewer:**

Name:

### DIFF
--- a/src/util/comparison.coffee
+++ b/src/util/comparison.coffee
@@ -55,6 +55,12 @@ module.exports.equivalent = equivalent = (a, b) ->
       return compareEveryItemInArrays(a, b, equivalent)
     when '[object Object]'
       return compareObjects(a, b, equivalent)
+    when '[object String]'
+      # Make sure b is also a string
+      if bClass == '[object String]'
+        # String equivalence is case-insensitive
+        a = a.toLowerCase()
+        b = b.toLowerCase()
 
   return equals a, b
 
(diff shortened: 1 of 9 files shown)


src/util/comparison.coffee
```diff
@@ -55,6 +55,12 @@ module.exports.equivalent = equivalent = (a, b) ->
       return compareEveryItemInArrays(a, b, equivalent)
     when '[object Object]'
       return compareObjects(a, b, equivalent)
+    when '[object String]'
+      # Make sure b is also a string
+      if bClass == '[object String]'
+        # String equivalence is case-insensitive
+        a = a.toLowerCase()
+        b = b.toLowerCase()
```
daco101: The spec notes that equivalence should be case and locale insensitive, but I think this will only achieve the first part. Looks like you should be able to use the js feature "localeCompare".

mayerm94: Isn't _not_ using localeCompare locale-insensitive behavior?

daco101: Hmm I dont think so. I think `define LocaleInsensitiveStrings: 'baréóa' ~ 'bArEoÄ'` should resolve to true.
`"baréóa".localeCompare("bArEoÄ", 'en', {sensitivity: 'base'})` is 0 (same), but `"baréóa".toLowerCase() == "bArEoÄ".toLowerCase()` is false

PR #58, opened 2018-09-20
Null if cant convert

losborne:
Some tests in this PR depend on the DateTime equality changes in https://github.com/cqframework/cql-execution/tree/bonnie_v1.3_equivalence_equality_updates

The main reason for the PR is to return null instead of throwing errors when things can't be converted, which is a change in the CQL spec.

Along the way, a couple of bugs were discovered that this PR also fixes.
1. ToTime wasn't parsing strings properly, so this PR ports over some of the DateTime.parse logic and adjusts it to be appropriate for Time.  Some helper functions were moved to util in order to facilitate this change.

2. Quantity arithmetic operators (+,-,.,/) were allowing quantities of different units to be operated on.  So `'1 m' + '1 foobar'` would be treated as `'1 m' + null`, which is `'1 m'`.  While the Quantity constructor guards against invalid UCUM units, so this case is not very likely, it was still possible.  The spec says that both Quantities must have the same units, so now `null` is returned if they don't.

https://jira.mitre.org/browse/BONNIE-1727

**Submitter:**
- [ ] Tests have been run locally and pass (some tests depend on bonnie_v1.3_equivalence_equality_updates, but they pass if that branch is included locally.  leaving unchecked to serve as marker for dependency)
- [x] Code coverage has not gone down and all code touched or added is covered. (all green lines in files changed are covered)
- [x] All dependent libraries are appropriately updated or have a corresponding PR related to this change N/A
- [ ] Regression test run and regressions logged to Jira

**Reviewer:**

Name:

### DIFF
--- a/src/datatypes/datetime.coffee
+++ b/src/datatypes/datetime.coffee
@@ -1,5 +1,9 @@
 { Uncertainty } = require './uncertainty'
-{ makeJsDate, jsDate } = require '../util/util'
+{ makeJsDate,
+  jsDate,
+  normalizeMillisecondsField,
+  normalizeMillisecondsFieldInString,
+  getTimezoneSeparatorFromString } = require '../util/util'
 moment = require 'moment'
 
 class DateTime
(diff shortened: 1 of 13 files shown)


jbradl11:
PR is being closed out due to duplicate PR: https://github.com/cqframework/cql-execution/pull/63

PR #59, opened 2018-09-20
Bonnie v1.3 object equivalence equality and datetime comparison updates

hossenlopp:

**Submitter:**
- [x] This pull request describes why these changes were made https://jira.mitre.org/browse/BONNIE-1702
- [x] ~~All dependent libraries are appropriately updated or have a corresponding PR related to this change~~ Will do later

**Reviewer:**

Name:


src/util/comparison.coffee
```diff
@@ -83,6 +83,8 @@ deepCompareKeysAndValues = (a, b, comparisonFunction) ->
   # Array.every() will only return true or false, so set a flag for if we should return null
   shouldReturnNull = false
   finalComparisonResult = aKeys.length is bKeys.length and aKeys.every (key) ->
+    # if both are null we should return true to satsify ignoring empty values in tuples
```
losborne: satsify -> satisfy

src/datatypes/datetime.coffee
```diff
@@ -676,6 +570,243 @@ DateTime.prototype.isSamePrecision = Date.prototype.isSamePrecision = (other) ->
...
+DateTime.prototype.before = Date.prototype.before = (other, precision) ->
+  if not((other.isDate) or (other.isDateTime))
+    return null
+  else if @.isDate and other.isDateTime
+    return @getDateTime().before(other, precision)
+  else if @.isDateTime and other.isDate
+    other = other.getDateTime()
+
+  if precision? && @constructor.FIELDS.indexOf(precision) < 0
+    throw new Error("Invalid precision: #{precision}")
```
losborne: We could also enumerate the valid precisions in this error message, which might be helpful.

test/datatypes/interval.coffee
```diff
@@ -739,11 +739,19 @@ describe 'DateTimeInterval.union', ->
 
     [x, y] = xy @dIvl.before
     # TODO: I don't know about these tests... doesn't make sense to me.
```
losborne: I think this TODO might be able to be removed

src/datatypes/datetime.coffee
hossenlopp: The CQL to ELM translator rejects invalid precisions. The only case that we will see this is if someone uploads a CQL 1.2 measure that uses `week` precision. Now that I think about it, I will create a test to make sure the right list is being used to reject the time level precisions on Date objects.

PR #63, opened 2018-09-25
Null if cant convert rebase

jbradl11:

Duplicate PR of Luke's: https://github.com/cqframework/cql-execution/pull/58

Some tests in this PR depend on the DateTime equality changes in https://github.com/cqframework/cql-execution/tree/bonnie_v1.3_equivalence_equality_updates

The main reason for the PR is to return null instead of throwing errors when things can't be converted, which is a change in the CQL spec.

Along the way, a couple of bugs were discovered that this PR also fixes.
1. ToTime wasn't parsing strings properly, so this PR ports over some of the DateTime.parse logic and adjusts it to be appropriate for Time.  Some helper functions were moved to util in order to facilitate this change.

2. Quantity arithmetic operators (+,-,.,/) were allowing quantities of different units to be operated on.  So `'1 m' + '1 foobar'` would be treated as `'1 m' + null`, which is `'1 m'`.  While the Quantity constructor guards against invalid UCUM units, so this case is not very likely, it was still possible.  The spec says that both Quantities must have the same units, so now `null` is returned if they don't.

**Submitter:**

**Reviewer:**

Name:


src/util/util.coffee
```diff
@@ -31,4 +31,26 @@ module.exports.makeJsDate = () ->
   # so using regular js here
   `new Date(...arguments)`
 
-module.exports.jsDate = Date
\ No newline at end of file
+module.exports.jsDate = Date
+module.exports.normalizeMillisecondsFieldInString = normalizeMillisecondsFieldInString = (string, matches) ->
+  msString = matches[14]
```
cmoesel: It seems a little dangerous that this util function relies on the specific form of the regex that produced `matches`.  If someone changes the regex in one of the places where this is used, and it changes the indexes in the `matches` array, it's easy to miss fixing this util function to match it.

I'd prefer that the calling code just pass in the ms string (`matches[14]`) directly rather than the whole `matches` array.

test/elm/convert/test.coffee
```diff
@@ -271,4 +328,4 @@ describe 'ToBoolean', ->
     should(@upperCaseT.exec(@ctx)).be.true
 
   it "should return true for F", ->
-    should(@upperCaseF.exec(@ctx)).be.false
\ No newline at end of file
+    should(@upperCaseF.exec(@ctx)).be.false
```
cmoesel: The test description is wrong for this one.

src/util/util.coffee
jbradl11: Good point, i'll update that now.

cmoesel:
Looks good.  Thanks!

cmoesel:
May require another rebase though, as I think others have pushed to `bonnie_v1.3_updates` since you made this PR.

PR #65, opened 2018-09-25
Bonnie v1.3 updates quantity null handling

daco101:
This PR adds functionality so that a Quantity with a null unit behaves the same as a Quantity with a default unit ('1') in all cases.
It also fixes it so that mismatched unit operations result in null, and adds tests for those cases.

It also fixes the cakefile to include test output of failed tests (e.g. yarn test will work as expected).

**Submitter:**

**Reviewer:**

Name:


src/elm/quantity.coffee
```diff
@@ -250,22 +252,26 @@ module.exports.parseQuantity = (str) ->
...
     # The units don't have to match (m and m^2), but must be convertable
     # we will choose the unit of a to be the unit we return
-    val = convert_value(b.value, b.unit, a.unit)
+    val = convert_value(b.value, b_unit, a_unit)
     return null unless val?
-    new Quantity({unit: a.unit, value: a.value + val})
+    new Quantity({unit: a_unit, value: a.value + val})
   else
-    a.copy?().add?(b.value, clean_unit(b.unit))
+    b_unit = if b instanceof Quantity then coalesceToOne(b.unit) else b.unit
```
cmoesel: I'm a little confused by the `else` clause here -- because it _also_ refers to `b.unit`.  If `b` is not a `Quantity`, why would we expect there to be a `b.unit` property?  Is there some case where a JSON object is passed in instead of a `Quantity`?

src/elm/quantity.coffee
```diff
@@ -250,22 +252,26 @@ module.exports.parseQuantity = (str) ->
...
+    a.copy?().add?(b.value, clean_unit(b_unit))
```
cmoesel: The only classes I can find with an `add` property are `DateTime` and `Date`.  Is that what this is intended to handle?  If so, should we be checking for it (via `else if` instead of just the `else`)?

src/elm/quantity.coffee
```diff
@@ -250,22 +252,26 @@ module.exports.parseQuantity = (str) ->
...
     # The units don't have to match (m and m^2), but must be convertable
-    val = convert_value(b.value, b.unit, a.unit)
+    val = convert_value(b.value, b_unit, a_unit)
     return null unless val?
-    new Quantity({unit: a.unit, value: a.value - val})
+    new Quantity({unit: a_unit, value: a.value - val})
   else
-    a.copy?().add?(b.value * -1 , clean_unit(b.unit))
+    b_unit = if b instanceof Quantity then coalesceToOne(b.unit) else b.unit
+    a.copy?().add?(b.value * -1 , clean_unit(b_unit))
```
cmoesel: Same comments from `doAddition` apply to `doSubstraction`

src/elm/quantity.coffee
```diff
@@ -274,3 +280,6 @@ module.exports.doDivision = (a,b) ->
 
 module.exports.doMultiplication = (a,b) ->
   if a instanceof Quantity then a.multiplyBy(b) else b.multiplyBy(a)
+
+coalesceToOne = (o) ->
+  if o? then o else '1'
```
cmoesel: As noted above, should we also be checking for empty string (`""`)?  I _think_ so?

src/elm/quantity.coffee
cmoesel: OK

src/elm/quantity.coffee
```diff
@@ -248,29 +242,32 @@ module.exports.parseQuantity = (str) ->
   else
     null
 
-module.exports.doAddition = (a,b) ->
+module.exports.doScaledAddition = (a,b,scaleForB) ->
```
cmoesel: A little nit-picky, but... given that this is really just a utility function used internally here by add and subtract, I'd prefer not to export it -- especially since its usage isn't 100% clear outside of the context of what you're doing here.

PR #66, opened 2018-09-25
Add ratio comparison

mayerm94:
Add ratio comparison to comparison.coffee so it calls ratio's already-existing equals function

**Submitter:**
- [] Code coverage has not gone down and all code touched or added is covered.

**Reviewer:**

Name:

### DIFF
--- a/src/elm/ratio.coffee
+++ b/src/elm/ratio.coffee
@@ -37,8 +37,7 @@ module.exports.Ratio = class Ratio extends Expression
       divided_this.equals(divided_other)
 
   equivalent: (other) ->
-    if other instanceof Ratio
-      @equals(other)
+    @equals(other)
 
 module.exports.createRatio = createRatio = (numerator, denominator) ->
   new Ratio({numerator: numerator, denominator: denominator})
--- a/src/elm/ratio.coffee
+++ b/src/elm/ratio.coffee
@@ -37,7 +37,10 @@ module.exports.Ratio = class Ratio extends Expression
       divided_this.equals(divided_other)
 
   equivalent: (other) ->
-    @equals(other)
+    equal = @equals(other)
+    if !equal?
+      return false
+    equal
 
 module.exports.createRatio = createRatio = (numerator, denominator) ->
   new Ratio({numerator: numerator, denominator: denominator})
--- a/src/elm/ratio.coffee
+++ b/src/elm/ratio.coffee
@@ -35,6 +35,8 @@ module.exports.Ratio = class Ratio extends Expression
       divided_this = @numerator.dividedBy(@denominator)
       divided_other = other.numerator.dividedBy(other.denominator)
       divided_this.equals(divided_other)
+    else
+      false
 
   equivalent: (other) ->
     equal = @equals(other)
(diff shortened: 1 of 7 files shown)


src/elm/ratio.coffee
```diff
@@ -36,5 +36,9 @@ module.exports.Ratio = class Ratio extends Expression
       divided_other = other.numerator.dividedBy(other.denominator)
       divided_this.equals(divided_other)
 
+  equivalent: (other) ->
+    if other instanceof Ratio
+      @equals(other)
```
jbradl11: This returns null if `other` is not a Ratio correct, was that what was decided?

jbradl11: Also, if this just calls the equals class in Ratio, that also does a check for `other instanceof Ratio` so I don't think you need a duplicate check here, unless we want to return a different "falsey" value for equivalence.

cmoesel:
OK.  Good news (for us).  CQL-to-ELM doesn't allow `<`, `<=`, `>`, `>=` with `Ratio`.  If you try it, you get an error like this:
```
<data.coffee#Less[16:20, 16:53]> Could not resolve call to operator Less with signature (System.Ratio,System.Ratio).
```

CQL-to-ELM _does_ support `=` and `~` though.  You may want to consider adding tests for Ratio in `test/elm/comparison/`.  For example:
```
define EqRatios: 10 'mg' : 2 'dL' = 15 'mg' : 3 'dL'
define UneqRatios: 10 'mg' : 2 'dL' = 15 'mg' : 4 'dL'
```
and
```
define EqRatios: 10 'mg' : 2 'dL' ~ 15 'mg' : 3 'dL'
define UneqRatios: 10 'mg' : 2 'dL' ~ 15 'mg' : 4 'dL'
```

test/elm/comparison/test.coffee
```diff
@@ -93,6 +93,12 @@ describe 'Equal', ->
   it 'should be null for 5 Cel = 40 m', ->
     should(@aLtB_Quantity_incompatible.exec(@ctx)).be.null()
 
+  it 'should be true for 10mg:2dL = 15mg:3dL', ->
+    @eqRatios.exec(@ctx).should.be.true()
+  
+  it 'should be true for 10mg:2dL = 15mg:4dL', ->
```
cmoesel: typo: it should be _false_

test/elm/comparison/test.coffee
```diff
@@ -197,7 +203,13 @@ describe 'Equivalent', ->
 
   it 'should be true for FOO ~ foo', ->
     @caseInsensitiveStrings.exec(@ctx).should.be.true()
+
+  it 'should be true for 10mg:2dL ~ 15mg:3dL', ->
+    @eqRatios.exec(@ctx).should.be.true()
   
+  it 'should be true for 10mg:2dL ~ 15mg:4dL', ->
```
cmoesel: typo: it should be _false_

test/elm/comparison/test.coffee
mayerm94: <facepalm> I shouldn't have let that go through

PR #67, opened 2018-09-27
Fix todate

daco101:
Fix ToDate to also accept a datetime, fixes Today to be a date.
This PR also fixes an issue with the date to datetime conversion (ToDateTime).

**Submitter:**

**Reviewer:**

Name:

### DIFF
--- a/src/elm/type.coffee
+++ b/src/elm/type.coffee
@@ -50,7 +50,7 @@ module.exports.ToDate = class ToDate extends Expression
 
   exec: (ctx) ->
     arg = @execArgs(ctx)
-    if (not arg?) or (arg == undefined)
+    if (not arg?)
       return null
     else if arg.isDateTime
       return arg.getDate()
@@ -63,7 +63,7 @@ module.exports.ToDateTime = class ToDateTime extends Expression
 
   exec: (ctx) ->
     arg = @execArgs(ctx)
-    if (not arg?) or (arg == undefined)
+    if (not arg?)
       return null
     else if arg.isDate
       return arg.getDateTime()
--- a/src/elm/type.coffee
+++ b/src/elm/type.coffee
@@ -50,7 +50,12 @@ module.exports.ToDate = class ToDate extends Expression
 
   exec: (ctx) ->
     arg = @execArgs(ctx)
-    if arg? and typeof arg != 'undefined' then Date.parse(arg.toString()) else null
+    if (not arg?) or (arg == undefined)
+      return null
+    else if arg.isDateTime
+      return arg.getDate()
+    else
+      return Date.parse(arg.toString())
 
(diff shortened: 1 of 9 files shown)


src/elm/type.coffee
```diff
@@ -50,7 +50,12 @@ module.exports.ToDate = class ToDate extends Expression
 
   exec: (ctx) ->
     arg = @execArgs(ctx)
-    if arg? and typeof arg != 'undefined' then Date.parse(arg.toString()) else null
+    if (not arg?) or (arg == undefined)
```
cmoesel: This is fine, but for future reference, I think `not arg?` will also be true for `undefined` -- so you don't need that second `arg == undefined`.

test/elm/convert/test.coffee
```diff
@@ -142,34 +142,49 @@ describe 'FromDateTime', ->
...
   @beforeEach ->
     setup @, data
 
-  it "should convert ToDate('2015-01-01') to DateTime", ->
+  it "should convert @2015-01-01 to DateTime", ->
     dateTime = @dateDateTime.exec(@ctx)
     dateTime.year.should.equal 2015
     dateTime.month.should.equal 1
     dateTime.day.should.equal 1
+    dateTime.timezoneOffset.should.equal @ctx.getTimezoneOffset()
```
cmoesel: This test should also check that all the time components (hour, minute, second, millisecond) are `0` (as specified in [Section 3.17](http://cql.hl7.org/09-b-cqlreference.html#todatetime) of the spec).

Looking at our current implementation ([here](https://github.com/cqframework/cql-execution/blob/master/src/elm/type.coffee#L63-L64)  and [here](https://github.com/cqframework/cql-execution/blob/master/src/datatypes/datetime.coffee#L503-L504)), I think that test would actually fail.  I know that's not code you touched in this PR, but you touched the test. ;-)

I actually should have caught this one before.

daco101: Good point. Fixed.

daco101:
I've made the fix in this PR and updated the description.

test/elm/convert/test.coffee
```diff
@@ -182,6 +186,7 @@ describe 'FromDate', ->
     date.year.should.equal 2015
     date.month.should.equal 1
     date.day.should.equal 1
+    should.not.exist(date[field]) for field in [ 'hour', 'minute', 'second', 'millisecond', 'timezoneOffset' ]
```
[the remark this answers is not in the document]
daco101: I've added the should.not.exist if that's what you mean?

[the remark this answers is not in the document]
cmoesel: Oh.  My bad.  I was thinking it was toDateTime, but it's toDate -- so you're right.

PR #75, opened 2018-12-19
Same as operator

zlister:

**Submitter:**

**Reviewer:**

Name:

### DIFF
--- a/src/datatypes/interval.coffee
+++ b/src/datatypes/interval.coffee
@@ -165,6 +165,11 @@ module.exports.Interval = class Interval
     else # ol is null
       null
 
+  sameAs: (other, precision) ->
+    if (@ == null or other == null) then return null
+    if (@low == null or @high == null or other.low == null or other.high == null) then return null
+    @low.sameAs(other.low, precision) and @high.sameAs(other.high, precision)
+
   equals: (other) ->
     if other instanceof Interval
       [a, b] = [@toClosed(), other.toClosed()]
--- a/src/elm/comparison.coffee
+++ b/src/elm/comparison.coffee
@@ -40,3 +40,12 @@ module.exports.GreaterOrEqual = class GreaterOrEqual extends Expression
     args = @execArgs(ctx).map (x) -> Uncertainty.from x
     return null unless args[0]? and args[1]?
     args[0].greaterThanOrEquals args[1]
+
+module.exports.SameAs = class SameAs extends Expression
+  constructor: (json) ->
+    super
+    @precision = json.precision
+
+  exec: (ctx) ->
+    [a, b] = @execArgs(ctx)
+    if a? and b? then a.sameAs(b, @precision?.toLowerCase()) else null
--- a/src/elm/datetime.coffee
+++ b/src/elm/datetime.coffee
@@ -113,15 +113,6 @@ module.exports.TimezoneFrom = class TimezoneFrom extends Expression
     date = @execArgs(ctx)
     if date? then date.timezoneOffset else null
 
-# module.exports.SameAs = class SameAs extends Expression
-#   constructor: (json) ->
-#     super
-#     @precision = json.precision
-
-#   exec: (ctx) ->
-#     [d1, d2] = @execArgs(ctx)
-#     if d1? and d2? then d1.sameAs(d2, @precision?.toLowerCase()) else null
-
 module.exports.SameOrAfter = class SameOrAfter extends Expression
   constructor: (json) ->
     super
--- a/src/elm/datetime.coffee
+++ b/src/elm/datetime.coffee
@@ -113,14 +113,14 @@ module.exports.TimezoneFrom = class TimezoneFrom extends Expression
     date = @execArgs(ctx)
     if date? then date.timezoneOffset else null
 
-module.exports.SameAs = class SameAs extends Expression
-  constructor: (json) ->
-    super
-    @precision = json.precision
+# module.exports.SameAs = class SameAs extends Expression
+#   constructor: (json) ->
+#     super
+#     @precision = json.precision
 
-  exec: (ctx) ->
-    [d1, d2] = @execArgs(ctx)
-    if d1? and d2? then d1.sameAs(d2, @precision?.toLowerCase()) else null
+#   exec: (ctx) ->
+#     [d1, d2] = @execArgs(ctx)
+#     if d1? and d2? then d1.sameAs(d2, @precision?.toLowerCase()) else null
 
 module.exports.SameOrAfter = class SameOrAfter extends Expression
   constructor: (json) ->
--- a/test/elm/interval/test.coffee
+++ b/test/elm/interval/test.coffee
@@ -2124,3 +2124,30 @@ describe 'DecimalIntervalExpand', ->
     # define BadPerMinute: expand { Interval(2.1, 4.1] } per 0.5 minute
     a = @badPerMinute.exec(@ctx)
     should.not.exist(a)
+
+describe 'SameAs', ->
+  @beforeEach ->
+    setup @, data
+
+  it 'returns null when both intervals values are null', ->
+    # define NullBoth: Interval[null,null] same as Interval[null,null]
+    a = @nullBoth.exec(@ctx)
+    should.not.exist(a)
+  
+  it 'returns true when both intervals are the same', ->
+    # define Equal: Interval[DateTime(2018,01,01), DateTime(2018,01,01)] same as Interval[DateTime(2018,01,01), DateTime(2018,01,01)]
+    @equal.exec(@ctx).should.be.true()
+
+  it 'returns false when both intervals are not the same', ->
+    # define NotEqual: Interval[DateTime(2018,01,01), DateTime(2018,01,01)] same as Interval[DateTime(2018,02,01), DateTime(2018,05,01)]
+    @notEqual.exec(@ctx).should.be.false()
+
+  it 'returns null when comparing date and datetime because precision is changed when converting date to datetime', ->
+    # define DateTimeAndDateComparisonEqual: Interval[DateTime(2018,01,01), DateTime(2018,01,01)] same as Interval[Date(2018,01,01), Date(2018,01,01)]
+    a = @dateTimeAndDateComparisonEqual.exec(@ctx)
+    should.not.exist(a)
+
+  it 'returns null when both intervals are null'. ->
+    # define NullIntervals: (null as Interval<DateTime>) same as (null as Interval<DateTime>)'
+    a = @nullIntervals.exec(@ctx)
+    should.not.exist(a)
--- a/test/elm/interval/test.coffee
+++ b/test/elm/interval/test.coffee
@@ -2147,7 +2147,88 @@ describe 'SameAs', ->
     a = @dateTimeAndDateComparisonEqual.exec(@ctx)
     should.not.exist(a)
 
-  it 'returns null when both intervals are null'. ->
+  it 'returns null when both intervals are null', ->
     # define NullIntervals: (null as Interval<DateTime>) same as (null as Interval<DateTime>)'
     a = @nullIntervals.exec(@ctx)
     should.not.exist(a)
+
+  it 'returns true when both Date intervals are the same', ->
+    # define DateIntervalComparisonSame: Interval[Date(2018,01,01), Date(2018,02,02)] same as Interval[Date(2018,01,01), Date(2018,02,02)]
+    @dateIntervalComparisonSame.exec(@ctx).should.be.true()
+
+  it 'returns false when Date intervals are not the same', ->  
+    # define DateIntervalComparisonNotSame: Interval[Date(2018,01,01), Date(2018,02,02)] same as Interval[Date(2018,01,01), Date(2018,02,01)]
+    @dateIntervalComparisonNotSame.exec(@ctx).should.be.false()
+   
+  it 'returns true when both Time intervals are the same', ->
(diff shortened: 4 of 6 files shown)


src/elm/comparison.coffee
```diff
@@ -40,3 +40,12 @@ module.exports.GreaterOrEqual = class GreaterOrEqual extends Expression
...
     args[0].greaterThanOrEquals args[1]
+
+module.exports.SameAs = class SameAs extends Expression
+  constructor: (json) ->
+    super
+    @precision = json.precision
+
+  exec: (ctx) ->
+    [a, b] = @execArgs(ctx)
+    if a? and b? then a.sameAs(b, @precision?.toLowerCase()) else null
```
cmoesel: I expect you moved `SameAs` from `datetime.coffee` to `comparison.coffee` because it now applies to both datetimes and intervals -- and comparison makes sense at a certain level.

That said, we've tried to organize these functions into files according to how the spec lists the operators in the appendix.  In the appendix, `SameAs` is not listed in the `Comparison` section (see: https://cql.hl7.org/09-b-cqlreference.html#comparison-operators-4) -- so this isn't where we'd normally put it.

If an operator is listed in more than one section (like in this case), we usually put it in the `overloaded.coffee` file.  For consistency, that's probably a better place for it than `comparison.coffee`.  Would you mind moving it there?

src/elm/datetime.coffee
```diff
@@ -113,14 +113,14 @@ module.exports.TimezoneFrom = class TimezoneFrom extends Expression
...
 
-module.exports.SameAs = class SameAs extends Expression
-  constructor: (json) ->
-    super
-    @precision = json.precision
-
-  exec: (ctx) ->
-    [d1, d2] = @execArgs(ctx)
-    if d1? and d2? then d1.sameAs(d2, @precision?.toLowerCase()) else null
+# module.exports.SameAs = class SameAs extends Expression
```
losborne: Should this commented-out code be removed?

test/elm/interval/test.coffee
```diff
@@ -2124,3 +2124,111 @@ describe 'DecimalIntervalExpand', ->
...
+    # define NotEqual: Interval[DateTime(2018,01,01), DateTime(2018,01,01)] same as Interval[DateTime(2018,02,01), DateTime(2018,05,01)]
+    @notEqual.exec(@ctx).should.be.false()
+
+  it 'returns null when comparing date and datetime because precision is changed when converting date to datetime', ->
+    # define DateTimeAndDateComparisonEqual: Interval[DateTime(2018,01,01), DateTime(2018,01,01)] same as Interval[Date(2018,01,01), Date(2018,01,01)]
+    a = @dateTimeAndDateComparisonEqual.exec(@ctx)
+    should.not.exist(a)
+
+  it 'returns null when both intervals are null', ->
+    # define NullIntervals: (null as Interval<DateTime>) same as (null as Interval<DateTime>)'
```
losborne: There's an extra `'` at the end of this line that can be removed.

src/datatypes/interval.coffee
```diff
@@ -165,6 +165,11 @@ module.exports.Interval = class Interval
     else # ol is null
       null
 
+  sameAs: (other, precision) ->
+    if (@ == null or other == null) then return null
```
losborne: This branch is not tested:
image

test/elm/interval/test.coffee
```diff
@@ -2124,3 +2124,111 @@ describe 'DecimalIntervalExpand', ->
     # define BadPerMinute: expand { Interval(2.1, 4.1] } per 0.5 minute
     a = @badPerMinute.exec(@ctx)
     should.not.exist(a)
+
+describe 'SameAs', ->
+  @beforeEach ->
+    setup @, data
+
+  it 'returns null when both intervals values are null', ->
+    # define NullBoth: Interval[null,null] same as Interval[null,null]
```
losborne: While I can see the benefit of looking at these while writing the tests, I would remove these comments showing the cql from this file, as if the CQL changes for some reason, these will become stale.

src/datatypes/interval.coffee
```diff
@@ -165,6 +165,11 @@ module.exports.Interval = class Interval
...
+    if (@low == null or @high == null or other.low == null or other.high == null) then return null
```
cmoesel: I don't think returning `null` if the low or high is `null` is quite right.  I think it's more complicated than that. According to the spec:
> **Same As**: The same-precision-as operator for intervals returns true if the two intervals start and end at the same value, using the semantics described in the `Start` and `End` operators to determine interval boundaries...

And if you look at the definitions of `Start` and `End`:

> **Start**: ... If the low boundary of the interval is closed and the low value of the interval is not null, this operator returns the low value of the interval. Otherwise, the result is the minimum value of the point type of the interval.

> **End**: ... If the high boundary of the interval is closed and the high value of the interval is not null, this operator returns the high value of the interval. Otherwise, the result is the maximum value of the point type of the interval.

So... in other words, if an interval boundary is _closed_ and its value is _null_, then it's treated like one of the min/max values (because its intentional is really negative or positive infinity).  In this respect, `Interval[@2018-01-01, null] same day as Interval[@2018-01-01, null]` should be `true`.

So... I think you really want to be looking at the calculated `Start` and `End` of the intervals, as the spec suggests.

cmoesel: I'm noticing now that it looks like maybe our `Start` and `End` implementations in `interval.coffee` are not right either, as they don't take into account whether the interval boundaries are open or closed.  We should probably fix that (although maybe that's a different pull request).

Looking at the code, I see that the `Interval` class has a `toClosed()` function on it.  It looks like that might actually take care of converting boundaries correctly and giving you an interval that is easier to compare on the low and high values.

Using `toClosed` on both intervals before comparing would also take care of converting open to closed boundaries, which would likely fix another use case that I think is currently broken.  In short, `Interval[@2018-01-01T00:00:00.0, @2019-01-01T00:00:00.0) same as Interval[@2018-01-01T00:00:00.0, @2018-12-31T23:59:59.999]` should also be `true` since the end boundaries are just two different ways of saying the same thing.

src/datatypes/interval.coffee
```diff
@@ -165,6 +165,11 @@ module.exports.Interval = class Interval
...
+    @low.sameAs(other.low, precision) and @high.sameAs(other.high, precision)
```
cmoesel: Would this work for intervals of numbers?  I think the intent of `Same As` is that it works on _all_ intervals (not just DateTimes).  I think this would fail on `Interval[0, 5] same as `Interval[0, 5]`.

src/datatypes/interval.coffee
zlister: I had a test for that, but null values are handled earlier in the execution. I have removed that line.

src/elm/comparison.coffee
zlister: Will do!

test/elm/interval/test.coffee
zlister: Decided to keep these because that is the format for the majority of the file

src/datatypes/datetime.coffee
```diff
@@ -841,6 +841,10 @@ DateTime.prototype.add = Date.prototype.add = (offset, field) ->
     for f in @constructor.FIELDS
       result[f] = null if not @[f]?
 
+  # To combat the conversion of the year 0 to 1900
```
losborne: What is the conversion of the year 0 to 1900?

src/datatypes/interval.coffee
```diff
@@ -165,6 +165,18 @@ module.exports.Interval = class Interval
     else # ol is null
       null
 
+  sameAs: (other, precision) ->
+    if ((@low == null and @high == null) or (other.low == null and other.high == null)) then return null
+    # pass in the opposite value to get the type since the checked value is null
+    else if @low == null then @low = minValueForInstance(@high)
```
losborne: It looks like this and the next line need a test case to cover the case when @low == null or other.low == null.

src/util/math.coffee
```diff
@@ -49,7 +49,7 @@ module.exports.successor = successor = (val) ->
       #not bothering with the max float test because javascript does not handle floats at the level
       #very well
       val + MIN_FLOAT_PRECISION_VALUE
-  else if val instanceof DateTime
+  else if val instanceof DateTime || val?.isDate
```
losborne: Should there be a similar case for the max Time value?

test/elm/interval/test.coffee
```diff
@@ -2124,3 +2167,159 @@ describe 'DecimalIntervalExpand', ->
...
+    a = @nullBoth.exec(@ctx)
+    should.not.exist(a)
```
losborne: It is better to use `should(a).be.null()` to test for null, as should.not.exist will pass for `undefined` as well.

test/elm/interval/test.coffee
```diff
@@ -2124,3 +2167,159 @@ describe 'DecimalIntervalExpand', ->
...
+
+  it 'returns null when one interval is null', ->
```
losborne: 'one interval is null' -> 'one interval's low and high are null' would be more precise

src/datatypes/datetime.coffee
zlister: When the date gets converted to a javascript Date object in the toJSDate() function on line 267, if the year is equal to 0, it returns a year of 1900. This affected when the min_date is used.

src/util/math.coffee
zlister: The Time values are converted to DateTime objects before they get to this point.

src/datatypes/datetime.coffee
losborne: Interesting--it looks like JS Date behaves differently for years 0-100, assuming they are in the 1900s.  It seems like there is a way to use setUTCFullYear to adjust for this 'feature'.  I think we should do that and move it to the toJSDateFunction rather than having it here.

src/util/math.coffee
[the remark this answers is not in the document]
cmoesel: I think `Date`s may need special logic that checks against `MAX_DATE_VALUE.getDate()` instead.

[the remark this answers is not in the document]
cmoesel: I won't leave any more comments on this, but note it would apply to three other places in this file as well.

src/datatypes/interval.coffee
```diff
@@ -217,6 +229,22 @@ module.exports.Interval = class Interval
...
+        return @low
+    return @toClosed().low
+
+  end: () ->
+    if !@high?
+      if @highClosed
+        return maxValueForInstance(@low)
+      else
+        return @high
+    return @toClosed().high
```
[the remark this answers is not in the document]
zlister: Yep it can! Thank you for pointing this out.

src/datatypes/interval.coffee
```diff
@@ -166,12 +166,16 @@ module.exports.Interval = class Interval
...
 
   sameAs: (other, precision) ->
-    if ((@low == null and @high == null) or (other.low == null and other.high == null)) then return null
-    # pass in the opposite value to get the type since the checked value is null
-    else if @low == null then @low = minValueForInstance(@high)
-    else if other.low == null then other.low = minValueForInstance(other.high)
-    else if @high == null then @high = maxValueForInstance(@low)
-    else if other.high == null then other.high = maxValueForInstance(other.low)
+    # Checks to see if any of the Intervals have a open, null boundary, returns null if true
+    if (@low == null and !@lowClosed) or (@high == null and !@highClosed) or (other.low == null and !other.lowClosed) or (other.high == null and !other.highClosed) then return null
```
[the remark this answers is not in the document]
cmoesel: It might be worth adding the two examples above as test cases in the unit tests too.

src/datatypes/interval.coffee
```diff
@@ -166,12 +166,16 @@ module.exports.Interval = class Interval
...
+
+    # For the special case where Interval[null,null] same as Interval[null,null] to return true
+    if @lowClosed and !@low? and @highClosed and !@high? then return other.lowClosed and !other.low? and other.highClosed and !other.high?
+
+    # For the special case where Interval[...] same as Interval[null,null] should return false
+    # This accounts for the inverse of the if statement above: where the second Interval is [null,null] and not the first Interval
+    if other.lowClosed and !other.low? and other.highClosed and !other.high? then return false
```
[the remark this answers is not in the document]
zlister: When other.start() or other.end() is called, it is assumed that if the value is null and closed, that the opposite boundary is not null and it is passed to the Min or Max value function. It uses the opposite boundary to determine the datatype that is to be returned. Since it is null, it just returns null. There really isn't a good way to pass in a value since this calculation is done inside the interval. It has no knowledge of the interval it is being compared to. That is why I have an if block to catch that case.

src/datatypes/interval.coffee
```diff
@@ -166,15 +166,40 @@ module.exports.Interval = class Interval
       null
 
   sameAs: (other, precision) ->
-    # Checks to see if any of the Intervals have a open, null boundary, returns null if true
-    if (@low == null and !@lowClosed) or (@high == null and !@highClosed) or (other.low == null and !other.lowClosed) or (other.high == null and !other.highClosed) then return null
+    # This large if and else if block handles the scenarios where there is an open ended null
+    # If both lows or highs exists, it can be determined that intervals are not Same As
```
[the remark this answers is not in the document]
zlister: Yea Luke and I spent a good chunk of time trying to come up with a better solution too. Definitely complex!

PR #80, opened 2019-02-14
maximum/mininum DateTime now uses execution timezoneOffset

losborne:
This PR includes a fix for `maximum DateTime` and `minimum DateTime`, so that they use the execution DateTime timezoneoffset.

Bonnie is currently using the branch `bonnie-v3.1.1`, which included this change for a patch release.

Internal Ticket: https://jira.mitre.org/browse/BONNIE-1897
External Ticket: https://oncprojectracking.healthit.gov/support/projects/BONNIE/issues/BONNIE-521

It also upadates handlebars (a dependency of nyc), which was being flagged by yarn audit.

**Submitter:**
- [x] All dependent libraries are appropriately updated or have a corresponding PR related to this change (there are JIRA sub-tasks to track that they will be completed)
- [x] Bonnie Frontend Regression Passes (HEAD~1 vs HEAD) (only expected opioid regressions due to unlocked cql logic.  used `min-max-datetime` bonnie branch.
- [x] Bonnie Frontend Excel Regression Passes (HEAD~1 vs HEAD)
- [x] Bonnie Backend Excel Regression Passes (HEAD~1 vs HEAD) (Only expected opioidd differences)

**Reviewer:**

Name: @daco101 

### DIFF
--- a/src/elm/arithmetic.coffee
+++ b/src/elm/arithmetic.coffee
@@ -208,7 +208,7 @@ module.exports.MinValue = class MinValue extends Expression
     if MIN_VALUES[@valueType]
       if @valueType == '{urn:hl7-org:elm-types:r1}DateTime'
         minDateTime = MIN_VALUES[@valueType].copy()
-        minDateTime.timezoneOffset = parseFloat(ctx.getTimezoneOffset())
+        minDateTime.timezoneOffset = ctx.getTimezoneOffset()
         return minDateTime
       else
         MIN_VALUES[@valueType]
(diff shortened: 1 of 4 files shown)


src/elm/arithmetic.coffee
```diff
@@ -206,7 +206,12 @@ module.exports.MinValue = class MinValue extends Expression
 
   exec: (ctx) ->
     if MIN_VALUES[@valueType]
-      MIN_VALUES[@valueType]
+      if @valueType == '{urn:hl7-org:elm-types:r1}DateTime'
+        minDateTime = MIN_VALUES[@valueType].copy()
+        minDateTime.timezoneOffset = parseFloat(ctx.getTimezoneOffset())
```
cmoesel: Why is `parseFloat` needed here (and in line 233 below)?  Are there use cases where `ctx.getTimezoneOffset()` might return a string?  I searched the rest of the codebase and as far as I can tell, we don't `parseFloat` offsets anywhere else.

losborne: Good catch.

It was used because I noticed that in Bonnie, a string was being passed around:

image

After some investigation, it looks like we are passing in '0' in Bonnie, where it should be 0.  I guess it was just working through the magic of JavaScript type conversion.

I'll take out the `ParseFloat` and modify Bonnie to properly pass in `0.0`.

losborne:
@cmoesel Bonnie testing and review has been completed--this should be good to merge.

PR #83, opened 2019-03-08
Set uncertainty values to null if they are codes

mayerm94:
Uncertainty low and high were being set to codes, causing codes to be compared to enumerables in instances where a value could be either a code or quantity (QDM.Result)

**Submitter:**

**Reviewer:**

Name:

**Reviewer:**

Name: @losborne 


src/datatypes/uncertainty.coffee
```diff
@@ -11,6 +11,10 @@ module.exports.Uncertainty = class Uncertainty
         # Uncertainties with different types probably shouldn't be supported.
         return false
       if typeof a.after is 'function' then a.after b else a > b
+    isCode = (val) ->
```
losborne: I think we should follow the recent pattern in `isQuantity` or `isDatetime` and add this to `src/datatypes/clinical.coffee`:

```coffeescript
  # Define a simple getter to allow type-checking of this class without instanceof
  # and in a way that survives minification (as opposed to checking constructor.name)
  Object.defineProperties @prototype,
    isCode:
      get: -> true
```

Then this logic would change to:
```coffeescript
if @low.isCode then @low = null
if @high.isCode then @high = null
```

cmoesel: @losborne makes a great point -- and I don't think it should be hard to implement (given he more or less included the implementation in his comment).  So... I'd suggest we follow his advice.

cmoesel: @mayerm94 mentioned in a private thread that we might also need to do the same for `ValueSet` and `Concept`.  I'm good with that as well!

src/datatypes/uncertainty.coffee
```diff
@@ -11,7 +11,11 @@ module.exports.Uncertainty = class Uncertainty
         # Uncertainties with different types probably shouldn't be supported.
         return false
       if typeof a.after is 'function' then a.after b else a > b
+    isNonEnumerable = (val) ->
+      val? and (val.isCode or val.isConcept or val.isValueSet)
     if typeof @high is 'undefined' then @high = @low
+    if isNonEnumerable(@low) then @low = null
+    if isNonEnumerable(@high) then @high = null
```
cmoesel: I hadn't thought of this before, but your unit tests made me think of it.  This could end up generating something like: `Uncertainty(1, null)`.  This concerns me because in the `lessThan` implementation (which everything else drives off of), `null` appears to have special meaning (it seems to be neg/pos infinity):
https://github.com/cqframework/cql-execution/blob/19775fd74c97c5d054e69f5aa6a050d9bb4fe450/src/datatypes/uncertainty.coffee#L48-L55

We don't want a code to be misinterpreted as an infinity.  I think it might be better that if either `@low` or `@high` are not enumerable, then we set _both_ to `null`.  In that case, I think `lessThan` will always return null.

What do you think of that?

PR #85, opened 2019-03-28
Add to ratio

losborne:
This should actually be 2 PRs but they are both small enough changes that I think it's doable in 1 PR.  If you disagree I can split them out.

Feature branch is temporary until cql-to-elm 1.4 is released, which will fix the skipped `null ~ null` test.

Updates include ToRatio implementation as well as changes from the new cql-to-elm snapshot.

Updated tests after Bryn's comment below, so this is no longer an issue for our tests.
```
New cql-to-elm snapshot caused failures in Date/DateTime equals/not equals tests. After fixing the Date/DateTime equals tests to match the spec, there are some differences (basically 1 underlying issue):

 1) Equal
      should be null for Date and DateTime equality with same year, month, hour and additional fields:
   `Date(2000, 3, 13) = DateTime(2000, 3, 13, 2, 4, 23)`
    AssertionError: expected false to be null

 2) Equal
      should be null for DateTime and Date equality with same year, month, hour and additional fields:
   `DateTime(2000, 3, 13, 2, 4, 23) = Date(2000, 3, 13)`
    AssertionError: expected false to be null

 3) NotEqual
      should be null for all Date and DateTime equality with same year, month, hour:
   `Date(2000, 3, 13) != DateTime(2000, 3, 13, 2, 4, 23)`
    AssertionError: expected true to be null

 4) NotEqual
      should be null for all DateTime and Date equality with same year, month, hour:
   `DateTime(2000, 3, 13, 2, 4, 23) != Date(2000, 3, 13)`
    AssertionError: expected true to be null


From the spec:

For Date, DateTime, and Time values, the comparison is performed by considering each precision in order, beginning with years (or hours for time values). If the values are the same, comparison proceeds to the next precision; if the values are different, the comparison stops and the result is false. If one input has a value for the precision and the other does not, the comparison stops and the result is null

https://github.com/HL7/cql/blob/master/spec/09-b-cqlreference.adoc#52-equal
```

**Submitter:**
- [x] Tests have been run locally and pass (4 failures unreleated to ToRatio)

**Reviewer:**

Name:


brynrhodes:
There's a caveat here that leads to some different behavior when comparing a _Date_ to a DateTime versus comparing DateTime to DateTime. I agree that the result for each of these would be null if we were starting out with DateTimes on both sides, but when we start with a Date on one side, the translator will implicitly convert that Date to a DateTime, and the ToDateTime(Date) function explicitly sets the Time components to zero: https://cql.hl7.org/2019May/09-b-cqlreference.html#todatetime
> For the Date overload, the result will be a DateTime with the time components set to zero, except for the timezone offset, which will be set to the timezone offset of the evaluation request timestamp.

So this CQL: 
    Date(2000, 3, 13) = DateTime(2000, 3, 13, 2, 4, 23)
Would actually show up in the ELM as:
    ToDateTime(Date(2000, 3, 13)) = DateTime(2000, 3, 13, 2, 4, 23)
Which is equivalent to:
    DateTime(2000, 3, 13, 0, 0, 0, 0) = DateTime(2000, 3, 13, 2, 4, 23)
Which I believe should compare false in 1.4.

I think it's worth considering whether this is really the behavior we want, since it results in some fairly inconsistent behaviors. The rationale was to make sure that it aligned with the Global.ToDate functionality.

src/elm/type.coffee
```diff
@@ -105,6 +106,26 @@ module.exports.ToQuantity = class ToQuantity extends Expression
...
+
+  exec: (ctx) ->
+    arg = @execArgs(ctx)
+    if arg? and typeof arg != 'undefined'
+      # Argument will be of form '<quantity>:<quantity>'
+      try
+        [numerator, denominator] = arg.toString().split(':').map((quantity) -> parseQuantity(quantity))
+      catch
+        # If the input string is not formatted correctly, or cannot be
+        # interpreted as a valid Ratio value, the result is null.
```
jbradl11: Should this comment say "as a valid Quantity value" ?

losborne:
```suggestion
        # interpreted as a valid Quantity value, the result is null.
```

losborne:
Updated tests to match behavior described by @brynrhodes.  They are now all passing.

PR #101, opened 2019-06-28
Normalize whitespace for equivalence

jbradl11:

https://jira.mitre.org/browse/BONNIE-2006

**Submitter:**

**Reviewer:**

Name: @losborne 

### DIFF
--- a/src/util/comparison.coffee
+++ b/src/util/comparison.coffee
@@ -65,8 +65,8 @@ module.exports.equivalent = equivalent = (a, b) ->
       # Make sure b is also a string
       if bClass == '[object String]'
         # String equivalence is case- and locale insensitive
-        a = a.replace(/\s+/g, ' ')
-        b = b.replace(/\s+/g, ' ')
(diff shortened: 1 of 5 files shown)


src/util/comparison.coffee
```diff
@@ -65,6 +65,8 @@ module.exports.equivalent = equivalent = (a, b) ->
       # Make sure b is also a string
       if bClass == '[object String]'
         # String equivalence is case- and locale insensitive
+        a = a.replace(/\s+/g, ' ')
```
losborne: I don't think this is quite right--this is saying that any number of whitespace characters is equivalent to any other number, but the spec says: ```Normalizing whitespace means that all whitespace characters are treated as equivalent, with whitespace characters as defined in the whitespace lexical category.```

So for example `"\t\t\t" ~ "\n\n\n"` should be true, but `"\t" ~ "\n\n" should be false because there are a different number of characters.

Removing the `+` from these 2 lines and adding some tests with varying number of characters should fix it.

PR #105, opened 2019-07-02
Expand semantics

losborne:
https://jira.mitre.org/browse/BONNIE-2010
Disregard this example in the spec--it is incorrect:
```
expand { Interval[10, 10] } per 0.1

{ Interval[10.0, 10.0], Interval[10.1, 10.1], ..., Interval[10.9, 10.9] }
```

**Submitter:**

**Reviewer:**

Name: @jbradl11 


src/elm/interval.coffee
```diff
@@ -345,27 +344,71 @@ module.exports.Expand = class Expand extends Expression
...
+    low = truncateDecimal(low, decimalPrecision)
+    high = truncateDecimal(high, decimalPrecision)
+
     return [] if low > high
-    results = (new dtivl.Interval(x, x + width, true, true) for x in [low..(high - width)] by per)
+    return [] unless low? and high?
+
+    # Deal with the weird case of a point interval expanding with more
+    # precision than it is defined by moving the high
+    if low is high
```
losborne: Check if potentially integers here

No way to tell if only 0s after the decimal, but can if there are nonzero decimal digits.

src/elm/interval.coffee
```diff
@@ -345,27 +344,54 @@ module.exports.Expand = class Expand extends Expression
...
+    current_low = low
+    results = []
+
+    return [] if perValue > (high - low + perUnitSize)
+    current_high = parseFloat((current_low + perValue - perUnitSize).toFixed(decimalPrecision))
+    intervalToAdd = new dtivl.Interval(current_low, current_high, true, true)
+    while intervalToAdd.high <= high
+      results.push(intervalToAdd)
+      # current_low = truncateDecimal(current_low + per.value, decimalPrecision)
+      # current_high = truncateDecimal(current_low + per.value - perUnitSize, decimalPrecision)
```
jbradl11: Should these lines be removed?

losborne: Yes, good catch

src/elm/interval.coffee
```diff
@@ -287,51 +287,50 @@ module.exports.Expand = class Expand extends Expression
...
+      current_high = current_low.add(per.value, per.unit).predecessor()
+      intervalToAdd = new dtivl.Interval(current_low, current_high, true, true)
+
     return results
 
+  truncateToPrecision: (value, unit) ->
+    # If interval boundaries are more precise than per quantity, truncate to
+    # the precision specified by the per
+    # If the per is more precise than the interval, fill out precision with
+    # minimum values so that the expansion can produce a more precise result
```
cmoesel: This comment:

> \# If the per is more precise than the interval, fill out precision with
> \# minimum values so that the expansion can produce a more precise result

seems out of place here, since it's not actually doing this in this code.

src/elm/interval.coffee
```diff
@@ -453,3 +477,33 @@ collapseIntervals = (intervals, perWidth) ->
     collapsedIntervals.push a
     collapsedIntervals
 
+getpointSize = (interval) ->
```
cmoesel: Is this function used?  As best I can tell, the one defined in the `Interval` datatype is used, but this one is not.

test/elm/interval/test.coffee
```diff
@@ -1660,18 +1659,9 @@ describe 'DateIntervalExpand', ->
     a = @nullBoth.exec(@ctx)
     should.not.exist(a)
 
-  it 'returns null when per is more precise than the interval ends', ->
+  it 'returns more precise intervals when per is more precise than the interval ends', ->
```
cmoesel: The description of this test is off.  I think it should be "returns empty list when..."

src/elm/interval.coffee
losborne: It was moved to `/datatype` and should be removed here. Good catch!

PR #109, opened 2019-07-17
Standards updates 2019 part 2

losborne:
More updates for CQL 1.4

The Expand PR didn't make it into this (yet?)

**Submitter:**

**Reviewer:**

Name:


src/elm/arithmetic.coffee
```diff
@@ -243,18 +274,38 @@ module.exports.Successor = class Successor extends Expression
...
   exec: (ctx) ->
     arg = @execArgs(ctx)
+    successor = null
     if (not arg?)
       null
     else
-      MathUtil.successor arg
+      try
+        # MathUtil.successor throws on underflow, and the exception is used in
+        # the logic for evaluating `meets`
```
cmoesel: If this is the case, shouldn't you be re-throwing the exception below?

src/elm/arithmetic.coffee
```diff
@@ -243,18 +274,38 @@ module.exports.Successor = class Successor extends Expression
...
+        successor = MathUtil.successor arg
+      catch e
+        if e instanceof MathUtil.OverFlowException
+          null
```
cmoesel: Again, based on the comment above, I don't know if we should actually be re-throwing this, but... as the code is now, I don't see the purpose of the `null` line above.  It's not the last line in the function, so it will not return.  And it is not assigned to anything.  So... it seems like a no-op?

src/elm/arithmetic.coffee
```diff
@@ -243,18 +274,38 @@ module.exports.Successor = class Successor extends Expression
...
       null
     else
-      MathUtil.predecessor arg
+      # MathUtil.predecessor throws on underflow, and the exception is used in
+      # the logic for evaluating `meets`
+      try
+        predecessor = MathUtil.predecessor arg
+      catch e
+        if e instanceof MathUtil.OverFlowException
+          null
```
cmoesel: Same comments as above apply here too.

src/elm/arithmetic.coffee
losborne: The comment was meant to indicate why we are `try..catch`ing rather than just adjusting `MathUtil.successor` to return `null` instead of throwing.  I'll update the comment to hopefully be clearer

src/elm/arithmetic.coffee
losborne: Good catch--it should be `return null` here.

src/elm/arithmetic.coffee
cmoesel: Oh right.  I get it now.  The other code counts on `MathUtil.successor` throwing, _not_ *this* code throwing...  Sorry for the mixup!

cmoesel:
I just rebased to cleanup the commits and remove the duplicate commits that must have somehow got in there through a merge.  I did a byte-by-byte diff between the original branch and my rebased branch and confirmed the contents are exactly the same.

PR #110, opened 2019-07-23
Return null for aggregate functions if quantity units are incompatible

losborne:
From the email thread concerning `Sum`, this PR returns `null` for aggregate functions when Quantity units are not compatible, e.g.:
```
Sum({9000 '/d', 200 'mg/d'}) = null
```

For `Min` and `Max`, the units are checked to be compatible, but the original unit is returned rather than the converted unit.

For `Interval` `start of` and `end of` operators it now ensures the values are in the context's timezoneOffset if they are maximum or minimum DateTime.

**Submitter:**

**Reviewer:**

Name: @cmoesel 

Name: @hossenlopp 


src/elm/aggregate.coffee
```diff
@@ -45,6 +49,14 @@ module.exports.Min = class Min extends AggregateExpression
     list = @source.execute(ctx)
     return null unless list?
     listWithoutNulls = removeNulls(list)
+
+    # Check for incompatible units and return null. We don't want to convert
+    # the units for Min/Max, so we throw away tue converted array if it succeeds
```
cmoesel: Small typo: `tue` -> `the`

src/elm/aggregate.coffee
```diff
@@ -60,6 +72,14 @@ module.exports.Max = class Max extends AggregateExpression
     items = @source.execute(ctx)
     return null unless items?
     listWithoutNulls = removeNulls(items)
+
+    # Check for incompatible units and return null. We don't want to convert
+    # the units for Min/Max, so we throw away tue converted array if it succeeds
```
cmoesel: Same typo: `tue` -> `the`

src/elm/aggregate.coffee
losborne: Good catch! Fixed this one and the one below just now

src/elm/interval.coffee
```diff
@@ -2,7 +2,7 @@
 { ThreeValuedLogic } = require '../datatypes/logic'
 { build } = require './builder'
 { Quantity, doAddition, doSubtraction, compare_units, convert_value } = require '../datatypes/quantity'
-{ successor, predecessor } = require '../util/math'
+{ successor, predecessor, maxValueForInstance, MAX_DATETIME_VALUE, minValueForInstance, MIN_DATETIME_VALUE } = require '../util/math'
```
cmoesel: Are you using `maxValueForInstance` or `minValueForInstance`?  I don't see any usages in the diff.

src/elm/interval.coffee
```diff
@@ -137,7 +137,12 @@ module.exports.Start = class Start extends Expression
   exec: (ctx) ->
     interval = @arg.execute(ctx)
     return null if !interval?
-    return interval.start()
+    start = interval.start()
+    # fix the timezoneOffset of minimum Datetime to match context offset
+    if start?.isDateTime && start.equals(MIN_DATETIME_VALUE)
+      start.timezoneOffset = ctx.getTimezoneOffset()
```
cmoesel: So... this wouldn't survive a round-trip, would it?  In other words, now `start` does *not* equal `MIN_DATETIME_VALUE` anymore -- and if we were to reconstruct an interval with it, it wouldn't be identified as the min/max anymore.  I guess that's OK?

src/elm/interval.coffee
hossenlopp: Good catch. These have been removed.

src/elm/interval.coffee
cmoesel: After discussing over Slack I get it now.  No problem.

PR #148, opened 2020-11-13
Fix params passing with dependent libraries

okeefm:
We recently discovered an issue with passing parameters to dependent libraries (aka libraries called by a Measure or other library). If those libraries defined a `default` value for the parameter, it was impossible to pass in another value. This PR fixes that, and adds tests.

**Submitter:**

**Reviewer:**

Name:

### DIFF
--- a/examples/browser/cql4browsers.js
+++ b/examples/browser/cql4browsers.js
@@ -10255,14 +10255,10 @@ var ParameterDef = /*#__PURE__*/function (_Expression) {
     value: function exec(ctx) {
       // If context parameters contains the name, return value.
       if (ctx && ctx.parameters[this.name] !== undefined) {
-        return ctx.parameters[this.name]; // If the parent context contains the name, return that value
-      } else if (ctx.getParentParameter(this.name) !== null) {
-        if (ctx.getParentParameter(this.name).execute !== undefined) {
-          return ctx.getParentParameter(this.name).execute(ctx);
-        } else {
-          return ctx.getParentParameter(this.name);
-        } // If default type exists, execute the default type
-
+        return ctx.parameters[this.name]; // If the parent context contains the name, return that
+      } else if (ctx.getParentParameter(this.name) !== undefined) {
+        var parentParam = ctx.getParentParameter(this.name);
+        return parentParam.default != null ? parentParam.default.execute(ctx) : parentParam; // If default type exists, execute the default type
       } else if (this.default != null) {
         return this.default != null ? this.default.execute(ctx) : undefined;
       }
--- a/examples/browser/cql4browsers.js
+++ b/examples/browser/cql4browsers.js
@@ -10256,12 +10256,15 @@ var ParameterDef = /*#__PURE__*/function (_Expression) {
       // If context parameters contains the name, return value.
       if (ctx && ctx.parameters[this.name] !== undefined) {
         return ctx.parameters[this.name]; // If the parent context contains the name, return that value
-      } else if (ctx.parent && ctx.getParentParameter(this.name) !== undefined) {
-        return ctx.getParentParameter(this.name); // If default type exists, execute the default type
+      } else if (ctx.getParentParameter(this.name) !== null) {
+        if (ctx.getParentParameter(this.name).execute !== undefined) {
+          return ctx.getParentParameter(this.name).execute(ctx);
+        } else {
+          return ctx.getParentParameter(this.name);
+        } // If default type exists, execute the default type
+
       } else if (this.default != null) {
-        return this.default != null ? this.default.execute(ctx) : undefined; // Else, if context and context's parent exist return the value of the parent's parameters with the given name.
-      } else {
-        return ctx.getParentParameter(this.name);
+        return this.default != null ? this.default.execute(ctx) : undefined;
       }
     }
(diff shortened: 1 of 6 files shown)


examples/browser/cql4browsers.js
```diff
@@ -10255,11 +10255,12 @@ var ParameterDef = /*#__PURE__*/function (_Expression) {
...
-        return ctx.parameters[this.name]; // If default type exists, execute the default type
+        return ctx.parameters[this.name]; // If the parent context contains the name, return that
+      } else if (ctx.getParentParameter(this.name) !== undefined) {
+        var parentParam = ctx.getParentParameter(this.name);
+        return parentParam.default != null ? parentParam.default.execute(ctx) : parentParam; // If default type exists, execute the default type
       } else if (this.default != null) {
-        return this.default != null ? this.default.execute(ctx) : undefined; // Else, if context and context's parent exist return the value of the parent's parameters with the given name.
-      } else {
-        return ctx.getParentParameter(this.name);
+        return this.default != null ? this.default.execute(ctx) : undefined;
```
cmoesel, line 10263: The check on `this.default != null` is unnecessary here since it's in a block that goes through a `this.default != null` gate.

src/elm/parameters.js
```diff
@@ -13,12 +13,13 @@ class ParameterDef extends Expression {
     // If context parameters contains the name, return value.
     if (ctx && ctx.parameters[this.name] !== undefined) {
       return ctx.parameters[this.name];
+      // If the parent context contains the name, return that
+    } else if (ctx.getParentParameter(this.name) !== undefined) {
+      const parentParam = ctx.getParentParameter(this.name);
+      return parentParam.default != null ? parentParam.default.execute(ctx) : parentParam;
       // If default type exists, execute the default type
     } else if (this.default != null) {
       return this.default != null ? this.default.execute(ctx) : undefined;
```
cmoesel, line 22: The check on `this.default != null` is unnecessary here since it's in a block that goes through a `this.default != null` gate.

okeefm: A good point...not sure why I would put that twice.

src/elm/parameters.js
```diff
@@ -13,12 +13,13 @@ class ParameterDef extends Expression {
     // If context parameters contains the name, return value.
     if (ctx && ctx.parameters[this.name] !== undefined) {
       return ctx.parameters[this.name];
+      // If the parent context contains the name, return that
+    } else if (ctx.getParentParameter(this.name) !== undefined) {
+      const parentParam = ctx.getParentParameter(this.name);
+      return parentParam.default != null ? parentParam.default.execute(ctx) : parentParam;
```
hossenlopp, line 19: Wouldn't this only take the default parameter of the parent library if it is defined, instead of what was passed in to the engine? The code seems to work despite this though.

okeefm: My best guess as to why this works is that the first case in this `if` handles when the param is passed into the engine and not defined in the parent lib.

okeefm: But I am by no means a `cql-execution` expert

cmoesel: Ha.  Yeah.  TBH, when I first saw this code, I thought the same thing as @hossenlopp.  And then came to the same conclusion as @okeefm (although I did not verify).  But it seems like that must be what is happening here.

hossenlopp: So `ctx.getParentParameter` seems to either return the value for that parameter OR a ParameterDef. That seems like an issue. In the debugger I have seen it return both. So I'm pretty confused about this whole thing.

cmoesel: Hmm... can you tell under what conditions it returns each?

cmoesel: I think that's why we have: `return parentParam.default != null ? parentParam.default.execute(ctx) : parentParam;`

In theory, every parameter definition should either have a passed in value or a default value (that's a CQL requirement -- no null parameter values).  If it's a passed in value, I think it would go through the first `if` or maybe to the "else" in the ternary on line 19.  If there is a default value, then it should return that default value per the ternary.  Which is probably why it works.

cmoesel: I agree that the way parameters work right now is confusing and a little difficult to follow.  But I'm not sure that now is the time to fix that, w/ others wanting a fix by COB.  As far as I can tell, this PR works in the standard use cases, and if there is a scenario it does not work, I expect it also doesn't work on master -- so this likely is not a regression.  But given it fixes the issue at hand, and we haven't yet identified a use case where it does not work, I'd lean toward accepting it.

cmoesel (approved):
Approving this PR as it fixes the issue at hand and does not (as far as we can tell) introduce any new regressions.

That said, @hossenlopp's excellent debugging skills have uncovered some issues that still remain, most notably that under certain circumstances, a `ParameterDefinition` may be returned where a _value_ would be expected. For example, this may happen when the main library defines a parameter w/ _no_ default, a dependency library redefines the parameter _with_ a default, and the library is executed w/ no parameter value passed in.  These issues will be addressed separately.

cmoesel:
Travis is stuck on queued -- but all the checks run successfully on my system -- so I am going to force merge this.

PR #214, opened 2021-01-20
Fix 197 and 198 for String Match and Split

### LINKED ISSUE #197: String Matches should only return true when it matches the entire input string
See spec-test: CqlStringOperatorsTest.Matches.MatchesWordsAndSpacesFalse

### LINKED ISSUE #198: String Split should return list w/ original string if separator is null
See spec-test: CqlStringOperatorsTest.Split.SplitABNull

kjmahalingam:

| **Test** | **Error** |
| --- | --- |
|CqlStringOperatorsTest.Matches.MatchesWordsAndSpacesFalse | Wrong answer (true vs false - engine is probably not looking for FULL match) |
| CqlStringOperatorsTest.Split.SplitABNull | Wrong answer (null vs {'a,b'}) |

Fixes #197
Fixes #198

**Submitter:**

**Reviewer:**

Name:

### DIFF
--- a/src/elm/string.js
+++ b/src/elm/string.js
@@ -49,11 +49,10 @@ class Split extends Expression {
   exec(ctx) {
     const stringToSplit = this.stringToSplit.execute(ctx);
     const separator = this.separator.execute(ctx);
-    if (stringToSplit == null || separator == null) {
-      return null;
-    } else {
+    if (stringToSplit && separator) {
       return stringToSplit.split(separator);
     }
+    return stringToSplit ? [stringToSplit] : null;
   }
 }
 
@@ -155,11 +154,9 @@ class Matches extends Expression {
     if (string == null || pattern == null) {
       return null;
     }
-    if (string.match(new RegExp(pattern))) {
-      return true;
-    } else {
-      return false;
-    }
+
+    const match = string.match(new RegExp(pattern));
+    return Boolean(match && match.length > 0 && match[0] === string);
   }
 }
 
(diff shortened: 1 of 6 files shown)


kjmahalingam:
To address #198 I had to update the `String` `Split` operation. It appears that there is similar functionality in the `SplitOnMatches` operation. Should this be updated to be similar to the new approach here, or should it remain as it is?

src/elm/string.js
```diff
@@ -155,11 +154,9 @@ class Matches extends Expression {
...
       return null;
     }
-    if (string.match(new RegExp(pattern))) {
-      return true;
-    } else {
-      return false;
-    }
+
+    const match = string.match(new RegExp(pattern));
+    return Boolean(match && match.length > 0 && match[0] === string);
```
cmoesel, line 159: Your code works, but I think the code below would work too and is maybe a little more straightforward?
```suggestion
    return (new RegExp('^' + pattern + '$')).test(string);
```

cmoesel (approved):
I'm approving because it works!  But I did suggest one possible simplification.  You can choose to do it or not.

src/elm/string.js
kjmahalingam: I'll test this out and verify it works, and if so then then I'll make the change!

kjmahalingam:
@cmoesel did you see my above comment about the `SplitOnMatches` operation? I'm curious to see what you think about that.

cmoesel:
> To address #198 I had to update the `String` `Split` operation. It appears that there is similar functionality in the `SplitOnMatches` operation. Should this be updated to be similar to the new approach here, or should it remain as it is?

Ah.  Good call.  Yeah, the CQL documentation [uses the same basic language](https://cql.hl7.org/09-b-cqlreference.html#splitonmatches) to describe `SplitOnMatches`, so it should probably work the same way as `Split`.  Thanks!

cmoesel (approved):
Looks..... awesome!

PR #216, opened 2021-01-28
List/Interval Fixes

### LINKED ISSUE #164: Fix List "Contains" and "In" operator when the list is null
If the list is `null` (e.g., `(null as List<String>) contains 'foo'`), the result should be `false`, but CQL Execution evaluates it to `null`.

See: https://cql.hl7.org/09-b-cqlreference.html#contains-1

### LINKED ISSUE #165: Fix "Distinct" operator when the list contains multiple nulls
When a list contains multiple `null`s, the `Distinct` operator should collapse them to a single `null`.  Currently it does not.

### LINKED ISSUE #166: Fix List "Except" operator when the 2nd argument is null
When the 2nd argument is null, it should be treated like an empty list (so 1st argument is returned as-is).  CQL Execution currently returns `null` if the 2nd argument is `null`.

See: https://cql.hl7.org/09-b-cqlreference.html#except-1

### LINKED ISSUE #167: Fix List "Interect" operator when both lists contain null
When both lists contain one or more nulls (each), the result should be a list with a single null.  Currently the result has no nulls.

See: https://cql.hl7.org/09-b-cqlreference.html#intersect-1

### LINKED ISSUE #168: Fix List "Length" operator when list is null
When the list is null, the length should be calculated as `0`, but currently we return `null`.

See: https://cql.hl7.org/09-b-cqlreference.html#length-1

### LINKED ISSUE #172: Fix "AllTrue" operator when the list contains null
AllTrue should only consider non-null elements.  Null elements should be ignored.  Currently, we return null if any elements are null.

See: https://cql.hl7.org/09-b-cqlreference.html#alltrue

### LINKED ISSUE #173: Fix "Count" operator when list is null
If the list is `null`, the count should be `0`.

See: https://cql.hl7.org/09-b-cqlreference.html#count

### LINKED ISSUE #194: Interval Contains should be false when Interval is null
See: https://cql.hl7.org/STU4/09-b-cqlreference.html#contains

See spec-test: CqlIntervalOperatorsTest.Contains.TestNullElement1

### LINKED ISSUE #195: In Interval should be false when Interval is null
See: https://cql.hl7.org/STU4/09-b-cqlreference.html#in

See spec-test: CqlIntervalOperatorsTest.In.TestInNull

kjmahalingam:

| **Test** | **Error** |
| --- | --- |
| CqlAggregateFunctionsTest.AllTrue.AllTrueNullFirst | Wrong answer (null should be ignored) |
| CqlIntervalOperatorsTest.Contains.TestNullElement1 | Wrong answer (null vs false) |
| CqlIntervalOperatorsTest.In.TestInNull | Wrong answer (true vs false) |
| CqlListOperatorsTest.Contains.ContainsNullLeft | Wrong answer (null vs false) |
| CqlListOperatorsTest.Distinct.DistinctNullNullNull | Wrong answer ({null, null, null} vs {null}) |
| CqlListOperatorsTest.Distinct.DistinctANullANull | Wrong answer ({'a'', null, null} vs {'a', null}) |
| CqlListOperatorsTest.Except.ExceptNullRight | Wrong answer (null vs {1, 4}) |
| CqlListOperatorsTest.In.In1Null | Wrong answer (null vs false) |
| CqlListOperatorsTest.Length.LengthNullList | Wrong answer (null vs 0) |

Fixes #168
Fixes #173
Fixes #172
Fixes #164
Fixes #194
Fixes #195
Fixes #165
Fixes #166
Fixes #167

**Submitter:**

**Reviewer:**

Name:


src/datatypes/interval.js
```diff
@@ -42,6 +42,9 @@ class Interval {
     if (this.highClosed && this.high != null && cmp.equals(this.high, item)) {
       return true;
     }
+    if (this.low == null && this.high == null) {
+      return false;
+    }
```
cmoesel, line 47: So... I think you did this to satisfy this test:
```
"TestInNull": Tuple{
    expression: 5 in Interval[null, null],
    output: false
  }
```
But now that I look at it, the test is just flat out wrong.  In CQL, null boundaries have a different meaning depending on if the boundary is closed or not.  A closed null boundary means "infinite", whereas an open null boundary means "unknown".  So, if we have `Interval[null, null]` (closed boundaries), then technically _everything_  should be contained by that.

The documentation for [Interval#in](http://build.fhir.org/ig/HL7/cql/09-b-cqlreference.html#in) even suggests this:
> For closed interval boundaries, if the interval boundary is null, the result of the boundary comparison is considered true.

So, I think we need to back out this change and for keep the test ignored, but add it to the section of ignored tests pending feedback from Bryn.  I'll reach out to him to make sure I'm not missing something.

src/elm/overloaded.js
```diff
@@ -74,8 +74,10 @@ class Except extends Expression {
 
   exec(ctx) {
     const [a, b] = this.execArgs(ctx);
-    if (a == null || b == null) {
+    if (a == null && b == null) {
       return null;
+    } else if (b == null) {
+      return a;
```
cmoesel, line 80: This behavior is actually specific only to lists.  From the spec on [List#Except](http://build.fhir.org/ig/HL7/cql/09-b-cqlreference.html#except-1):
> If the second argument is null, the operation is performed as though the second argument was an empty list.

BUT... from the spec on [Interval#except](http://build.fhir.org/ig/HL7/cql/09-b-cqlreference.html#except):
> If either argument is null, the result is null.

So you need to branch based on `typeIsArray(a)` to decide how a null `b` should be handled.

test/spec-tests/cql/CqlIntervalOperatorsTest.cql
```diff
@@ -515,11 +513,9 @@ define "Except": Tuple{
 
 define "In": Tuple{
   "TestInNull": Tuple{
-    skipped: 'Wrong answer (true vs false)'
```
cmoesel, line 518: This is the test I think we need to continue ignoring but check in w/ Bryn about because I think the answer _should_ be true.

cmoesel (changes requested):
This is good!  I just have a couple of minor tweaks that I think need to be made.  See my individual comments...

src/datatypes/interval.js
cmoesel: Update: Bryn did confirm that the test is wrong.  The answer should be "true" and the name of the test should actually be "TestInNullBoundaries".  So can you just change the `false` to `true` in this test (and still back out the change in the logic so that the test passes)?

test/spec-tests/cql/CqlIntervalOperatorsTest.cql
cmoesel: As I noted above, Bryn confirmed that the test is wrong.  Can you please fix it according to the details in my other comment?

kjmahalingam: I am working on making this change; does anything need to be changed in the `CQLIntervalOperatorsTest.xml` file, or should this stay true to the contents of the spec?

kjmahalingam: Upon further testing, I've come to learn that the XML file is the one that requires the change in order for the build of CQL and ELM to work! Updating that accordingly, though I don't know if we need to be concerned about this file diverging from the spec.

kjmahalingam:
@cmoesel I believe all of your comments should be addressed!

test/spec-tests/cql/CqlIntervalOperatorsTest.cql
cmoesel: Right.  While we want to try to stay as close to the spec test files as possible, we've already made modifications where tests are blatantly wrong.  In those cases, I've also submitted a pull request to the CQL spec repo with the same changes (so if/when that PR is accepted, we'll be in alignment again).

For this specific change, I did not submit a PR to the CQL spec repo, but Bryn said he would fix it himself.

cmoesel (approved):
Thanks for addressing the issues.  This looks great!

PR #217, opened 2021-01-29
Replace ucum lib

### LINKED ISSUE #186: Arithmetic on Quantities should use smallest unit in result
When arithmetic is performed on quantities w/ different (but compatible) units, the smallest unit should be used for the result.  This might actually be OK for the standard cases.

Where we need to look at it is for multiply, where it seems that two same units might result in a larger squared unit (e.g., 'cm' * 'cm' results in 'm^2' instead of 'cm^2'.

See spec test: CqlArithmeticFunctionsTest.Multiply.Multiply1CMBy2CM

### LINKED ISSUE #190: Equivalence not working correctly for some quantities
For example, `1'cm' ~ 0.01'm'` should be equivalent, but is not evaluated as such.

See spec-test: CqlComparisonOperatorsTest.Equivalent.EquivEqCM1M01

cmoesel:
The ucum lib was quite old and unmaintained. It also had some ugly bits and we needed to run it through babel for it to work in browsers.  This PR replaces the old UCUM library w/ a UCUM library from NLM.  It also fixes at least one quantity bug we had.

As part of this, I also extracted all of the unit-based capabilities into their own module.  I had to reimplement some of the math stuff, particulary getting the correct result units when multiplying and dividing.  I think (hope) the implementation is actually better now than it was.  It is at least more understandable than the previous implementation.

Another added benefit is that it reduces the filesize of the overall library, so that's nice.

For all this work, this only really fixes one test from the spec (oh well):
| **Test** | **Error** |
| --- | --- |
| CqlArithmeticFunctionsTest.Multiply.Multiply1CMBy2CM | Wrong answer (.0002 m^2 vs 2 cm^2 - answer should be in terms of smallest unit) |
| CqlComparisonOperatorsTest.Equivalent.EquivEqCM1M01 | Wrong answer (false vs true) |

Fixes #186 
Fixes #190

**Submitter:**
- [?] Code coverage has not gone down and all code touched or added is covered.

**Reviewer:**

Name:

### DIFF
--- a/bin/browserify.js
+++ b/bin/browserify.js
@@ -7,17 +7,11 @@
 const path = require('path');
 const fs = require('fs');
 const browserify = require('browserify');
-const babelify = require('babelify');
 
 console.log('Browserifing cql4browsers...');
 const c4bpath = path.join(__dirname, '..', 'examples', 'browser', 'cql4browsers.js');
 const outputJsFile = fs.createWriteStream(c4bpath);
 browserify(path.join(__dirname, '..', 'examples', 'browser', 'simple-browser-support.js'))
-  .transform(babelify, {
-    global: true,
-    only: [/node_modules\/ucum\//],
-    presets: [['@babel/preset-env', { modules: false }]]
-  })
   .bundle()
   .pipe(outputJsFile);
 outputJsFile.on('finish', () => {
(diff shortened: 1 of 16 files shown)


kjmahalingam (approved):
Awesome work! I looked at this whole thing, and it's super comprehensive and seems to make things a lot nicer. Well done!

PR #218, opened 2021-02-03
Fix interval handling on null boundaries

### LINKED ISSUE #196: ProperlyIncludes/ProperlyIncludedIn wrong when using infinite interval
`Interval[null, null] properly includes Interval[1, 10]` should be true since the first interval is definitely larger than the second and contains the second -- but it results in `null` in the engine.  Same for reverse case w/ ProperlyIncludedIn.

_NOTE: The tests were also wrong, as they indicated the answer should be false!_

See spec-tests:
* CqlIntervalOperatorsTest.ProperlyIncludes.TestProperlyIncludesNull
* CqlIntervalOperatorsTest.ProperlyIncludedIn.TestProperlyIncludedInNull

cmoesel:
Various changes to address handling of intervals with null boundaries, especially with double-null boundaries.  Also cleans up a few other things related to `isTime()` that I stumbled upon along the way.  Changes:
- maintain the interval point type, considering casts when necessary
- use point type in calculations when necessary
- toClosed() function should not close null endpoints since that changes their meaning
- support interval construction when isLowExpression or isHighExpression properties are used
- fix incorrect uses of isTime to isTime()
- remove unnecessary uses of isTime()
- since Time is represented using a DateTime class, always check isTime() first when determining exact type

This fixes the following two test cases:
| **Test** | **Error** |
| --- | --- |
| CqlIntervalOperatorsTest.ProperlyIncludes.TestProperlyIncludesNull | Wrong answer (null vs false) |
| CqlIntervalOperatorsTest.ProperlyIncludedIn.TestProperlyIncludedInNull | Wrong answer (null vs false) |

_NOTE: The tests above were also renamed to better indicate what they are testing.  The new names are `NullBoundariesProperlyIncludesIntegerInterval` and `IntegerIntervalProperlyIncludedInNullBoundaries`.  In addition, after consulting w/ Bryn, I confirmed that the test cases themselves were actually wrong and the expected output should be `true`._

The following two test cases were not ignored before, but need to be ignored now.  This is because the CQL to ELM translator translates them wrong.  I've filed a bug for this (cqframework/clinical_quality_language#596), but until it is fixed, these test cases must be ignored:
| **Test** | **Error** |
| --- | --- |
| CqlIntervalOperatorsTest.After.TestAfterNull | Converts null to Interval[null,null] instead of Interval(null,null) |
| CqlIntervalOperatorsTest.Before.TestBeforeNull | Converts null to Interval[null,null] instead of Interval(null,null) |

Fixes #196 

**Submitter:**

**Reviewer:**

Name:

### DIFF
--- a/examples/browser/cql4browsers.js
+++ b/examples/browser/cql4browsers.js
@@ -2072,18 +2072,22 @@ var _require4 = require('../util/math'),
     successor = _require4.successor,
     predecessor = _require4.predecessor,
     maxValueForInstance = _require4.maxValueForInstance,
-    minValueForInstance = _require4.minValueForInstance;
+    minValueForInstance = _require4.minValueForInstance,
+    maxValueForType = _require4.maxValueForType,
+    minValueForType = _require4.minValueForType;
 
 var cmp = require('../util/comparison');
 
 var Interval = /*#__PURE__*/function () {
-  function Interval(low, high, lowClosed, highClosed) {
+  function Interval(low, high, lowClosed, highClosed, defaultPointType) {
     _classCallCheck(this, Interval);
 
     this.low = low;
     this.high = high;
     this.lowClosed = lowClosed != null ? lowClosed : true;
-    this.highClosed = highClosed != null ? highClosed : true;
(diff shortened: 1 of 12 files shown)


kjmahalingam (approved):
Nice work!

PR #222, opened 2021-02-10
Uncertainty Arithmetic

### LINKED ISSUE #177: Support arithmetic for uncertainties
See: https://cql.hl7.org/05-languagesemantics.html#arithmetic-operators-2

kjmahalingam:

| **Test** | **Error** |
| --- | --- |
| "CqlDateTimeOperatorsTest.Uncertainty tests.DateTimeDurationBetweenUncertainAdd" | Adding Uncertainties not supported |
| "CqlDateTimeOperatorsTest.Uncertainty tests.DateTimeDurationBetweenUncertainSubtract" | Subtracting Uncertainties not supported |
| "CqlDateTimeOperatorsTest.Uncertainty tests.DateTimeDurationBetweenUncertainMultiply" | Multiplying Uncertainties not supported |

Fixes #177 

**Submitter:**

**Reviewer:**

Name:

### DIFF
--- a/src/elm/arithmetic.js
+++ b/src/elm/arithmetic.js
@@ -8,6 +8,7 @@ const {
   doMultiplication,
   doDivision
 } = require('../datatypes/quantity');
+const { Uncertainty } = require('../datatypes/uncertainty');
 
 class Add extends Expression {
   constructor(json) {
@@ -23,6 +24,8 @@ class Add extends Expression {
     const sum = args.reduce((x, y) => {
       if (x.isQuantity || x.isDateTime || x.isDate || (x.isTime && x.isTime())) {
         return doAddition(x, y);
+      } else if (x.isUncertainty && y.isUncertainty) {
+        return new Uncertainty(x.low + y.low, x.high + y.high);
       } else {
         return x + y;
       }
@@ -49,6 +52,8 @@ class Subtract extends Expression {
     const difference = args.reduce((x, y) => {
       if (x.isQuantity || x.isDateTime || x.isDate) {
         return doSubtraction(x, y);
+      } else if (x.isUncertainty && y.isUncertainty) {
+        return new Uncertainty(x.low - y.low, x.high - y.high);
       } else {
         return x - y;
       }
@@ -75,6 +80,8 @@ class Multiply extends Expression {
     const product = args.reduce((x, y) => {
       if (x.isQuantity || y.isQuantity) {
         return doMultiplication(x, y);
+      } else if (x.isUncertainty && y.isUncertainty) {
+        return new Uncertainty(x.low * y.low, x.high * y.high);
       } else {
         return x * y;
       }
--- a/src/elm/arithmetic.js
+++ b/src/elm/arithmetic.js
@@ -22,10 +22,25 @@ class Add extends Expression {
     }
 
     const sum = args.reduce((x, y) => {
+      if (x.isUncertainty && !y.isUncertainty) {
+        y = new Uncertainty(y, y);
+      } else if (y.isUncertainty && !x.isUncertainty) {
+        x = new Uncertainty(x, x);
+      }
+
       if (x.isQuantity || x.isDateTime || x.isDate || (x.isTime && x.isTime())) {
         return doAddition(x, y);
       } else if (x.isUncertainty && y.isUncertainty) {
-        return new Uncertainty(x.low + y.low, x.high + y.high);
+        if (
+          x.low.isQuantity ||
+          x.low.isDateTime ||
+          x.low.isDate ||
+          (x.low.isTime && x.low.isTime())
+        ) {
+          return new Uncertainty(doAddition(x.low, y.low), doAddition(x.high, y.high));
+        } else {
+          return new Uncertainty(x.low + y.low, x.high + y.high);
+        }
       } else {
         return x + y;
       }
@@ -50,10 +65,20 @@ class Subtract extends Expression {
     }
 
     const difference = args.reduce((x, y) => {
+      if (x.isUncertainty && !y.isUncertainty) {
+        y = new Uncertainty(y, y);
+      } else if (y.isUncertainty && !x.isUncertainty) {
+        x = new Uncertainty(x, x);
+      }
+
       if (x.isQuantity || x.isDateTime || x.isDate) {
         return doSubtraction(x, y);
       } else if (x.isUncertainty && y.isUncertainty) {
-        return new Uncertainty(x.low - y.low, x.high - y.high);
+        if (x.low.isQuantity || x.low.isDateTime || x.low.isDate) {
+          return new Uncertainty(doSubtraction(x.low, y.high), doSubtraction(x.high, y.low));
+        } else {
+          return new Uncertainty(x.low - y.high, x.high - y.low);
+        }
       } else {
         return x - y;
       }
@@ -78,10 +103,20 @@ class Multiply extends Expression {
     }
 
     const product = args.reduce((x, y) => {
+      if (x.isUncertainty && !y.isUncertainty) {
+        y = new Uncertainty(y, y);
+      } else if (y.isUncertainty && !x.isUncertainty) {
+        x = new Uncertainty(x, x);
+      }
+
       if (x.isQuantity || y.isQuantity) {
         return doMultiplication(x, y);
       } else if (x.isUncertainty && y.isUncertainty) {
-        return new Uncertainty(x.low * y.low, x.high * y.high);
+        if (x.low.isQuantity) {
+          return new Uncertainty(doMultiplication(x.low, y.low), doMultiplication(x.high, y.high));
+        } else {
+          return new Uncertainty(x.low * y.low, x.high * y.high);
+        }
       } else {
         return x * y;
       }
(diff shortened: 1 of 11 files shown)


src/elm/arithmetic.js
```diff
@@ -75,6 +80,8 @@ class Multiply extends Expression {
     const product = args.reduce((x, y) => {
       if (x.isQuantity || y.isQuantity) {
         return doMultiplication(x, y);
+      } else if (x.isUncertainty && y.isUncertainty) {
+        return new Uncertainty(x.low * y.low, x.high * y.high);
```
cmoesel, line 84: What do you think about divide?  I guess that would work a little different -- because the same approach does not apply. For example, `Uncertainty(5,10) / Uncertainty(5,10)` is definitely _not_ `Uncertainty(5/5, 10/10)` -- because `Uncertainty(1, 1)` is `1`.  But I think if we think about it from lowest possible value to highest possible value, then maybe it would be `Uncertainty(5/10, 10/5)`?  Or in otherwords, `Uncertainty A / Uncertainty B` is `Uncertainty(A.low/B.high, A.high/B.low)`? Does that make sense?  Should we implement it?

src/elm/arithmetic.js
```diff
@@ -49,6 +52,8 @@ class Subtract extends Expression {
     const difference = args.reduce((x, y) => {
       if (x.isQuantity || x.isDateTime || x.isDate) {
         return doSubtraction(x, y);
+      } else if (x.isUncertainty && y.isUncertainty) {
+        return new Uncertainty(x.low - y.low, x.high - y.high);
```
cmoesel, line 56: Actually, this is kind of like the division problem I discussed below.  I don't think it quite works like you've implemented here because something like `Uncertainty(5,10) - Uncertainty(5,10)` is not `Uncertainty(5-5, 10-10)` (which is the same as `0`).  If you think about it logically, we're saying:
> What is _some number between 5 and 10_ minus _some other number between 5 and 10_?

Well, we know the lowest possible result would be if the first number was actually `5` and the second number was actually `10`-- and the highest possible result would be if the first number was actually `10` and the second number was actually `5`.  So I think `Uncertainty(5,10) - Uncertainty(5,10)` should result in `Uncertainty(-5, 5)`, right?  Or more generally, `Uncertainty A - Uncertainty B` is `Uncertainty(A.low - B.high, A.high - B.low)`.

This makes sense if you convert it to addition too.  Another way to think of `- Uncertainty(5, 10)` is `+ Uncertainty(-10, -5)` (but note we had to switch the order of the numbers since `-10` < `-5`).

Anyway, hopefully my babbling makes some amount of sense.

src/elm/arithmetic.js
```diff
@@ -23,6 +24,8 @@ class Add extends Expression {
     const sum = args.reduce((x, y) => {
       if (x.isQuantity || x.isDateTime || x.isDate || (x.isTime && x.isTime())) {
         return doAddition(x, y);
+      } else if (x.isUncertainty && y.isUncertainty) {
+        return new Uncertainty(x.low + y.low, x.high + y.high);
```
cmoesel, line 28: I think we also need to test what _type_ of uncertainty this is.  If it is an uncertainty with dates as the low/high, for example, this would not work.  So you probably need to check the point types too.  To cover the bases, I guess it might look like:
```
} else if (x.isUncertainty && y.isUncertainty) {
  if (x.low.isQuantity || x.low.isDateTime || x.low.isDate || (x.low.isTime && x.low.isTime()) {
    return new Uncertainty(doAddition(x.low, y.low), doAddition(x.high, y.high));
  } else {
    return new Uncertainty(x.low + y.low, x.high + y.high);
  }
} else {
```

cmoesel: And this same approach would likely be appropriate for the others.

src/elm/arithmetic.js
```diff
@@ -23,6 +24,8 @@ class Add extends Expression {
     const sum = args.reduce((x, y) => {
```
cmoesel, line 24: It used to be that `Add` (and some of the other basic math operations) supported a variable number of arguments, hence the use of `reduce`.  But now, `Add` is a `BinaryExpression` -- meaning it only accepts two arguments.  So if it simplifies things, just be aware that you can decide not to use reduce anymore if you want.

cmoesel: Up to you though.

cmoesel (changes requested):
This is a good start, but it occurs to me that we probably should have discussed this in more detail earlier.  Uncertainties are a tricky concept in CQL and not documented as well as they could be.  Anyway, a few thoughts... 

1. I think that the implementation of subtraction is wrong (when I think about what uncertainties represent logically).  So... if you agree with my logic, does that mean the test is actually wrong?

2. I think we should implement `divide`.  The spec isn't clear, but it says [basic arithmetic operators are supported](http://build.fhir.org/ig/HL7/cql/05-languagesemantics.html#arithmetic-operators-2).  I think `divide` fits into that category.

3. I think we also should support these operations between an uncertainty and a finite value. E.g., `Uncertainty(5, 10) + 2` should be `Uncertainty(7, 12)`.  You can always convert a finite value to a single-point uncertainty to do this (i.e., treat `2` as `Uncertainty(2, 2)`.

4. There are a few other bonus comments/suggestions in there for you.  I won't repeat them here!

kjmahalingam:
> This is a good start, but it occurs to me that we probably should have discussed this in more detail earlier. Uncertainties are a tricky concept in CQL and not documented as well as they could be. Anyway, a few thoughts...
> 
> 1. I think that the implementation of subtraction is wrong (when I think about what uncertainties represent logically).  So... if you agree with my logic, does that mean the test is actually wrong?
> 2. I think we should implement `divide`.  The spec isn't clear, but it says [basic arithmetic operators are supported](http://build.fhir.org/ig/HL7/cql/05-languagesemantics.html#arithmetic-operators-2).  I think `divide` fits into that category.
> 3. I think we also should support these operations between an uncertainty and a finite value. E.g., `Uncertainty(5, 10) + 2` should be `Uncertainty(7, 12)`.  You can always convert a finite value to a single-point uncertainty to do this (i.e., treat `2` as `Uncertainty(2, 2)`.
> 4. There are a few other bonus comments/suggestions in there for you.  I won't repeat them here!

This is excellent input. I will respond in order here.

1. Your logic behind the subtraction operation makes perfect sense. I can implement it as such, but that does mean that, yes, the spec test is wrong. I can edit the spec test to actually behave like we would expect it to.
2. I think division operating similarly to subtraction seems right to me. There is not a spec test for this. I will write some additional tests for the uncertainty operations in the handwritten tests as well.
3. Yes, this is a good idea. Does that mean that instead of checking for `x.isUncertainty && y.isUncertainty` I should be checking for `x.isUncertainty || y.isUncertainty` and operate appropriately? I ask because I assume that, for example, `Uncertainty(5, 10) + 2` and `2 + Uncertainty(5, 10)` should be equally valid?
4. Your points are acknowledged and will be addressed!

cmoesel:
>     3\. Yes, this is a good idea. Does that mean that instead of checking for `x.isUncertainty && y.isUncertainty` I should be checking for `x.isUncertainty || y.isUncertainty` and operate appropriately? I ask because I assume that, for example, `Uncertainty(5, 10) + 2` and `2 + Uncertainty(5, 10)` should be equally valid?

Ah, good point.  Yes, I think so!  Sorry it is getting so complex!

src/elm/type.js
```diff
@@ -118,9 +119,15 @@ class ToDecimal extends Expression {
   exec(ctx) {
     const arg = this.execArgs(ctx);
     if (arg != null) {
-      const decimal = limitDecimalPrecision(parseFloat(arg.toString()));
-      if (isValidDecimal(decimal)) {
-        return decimal;
+      if (arg.isUncertainty) {
```
kjmahalingam, line 122: I don't know how I feel about this change. It was required in order to get division to work correctly. This is because unlike the other basic arithmetic operators, the arguments of the division operator are processed as though they need to go through `ToDecimal`.
Technically all is working now since I've handled the case of an uncertainty here, but it feels weird to have an uncertainty check in only this one operation. Do you feel like it needs to be added in any of the other type operations?

kjmahalingam:
@cmoesel All feedback is addressed. I do have one comment I just made inline [here](https://github.com/cqframework/cql-execution/pull/222#discussion_r577132974) that I want your input on, but pending that everything else should be all set.

src/elm/type.js
cmoesel: This is an interesting one.  Looking at the spec, `ToDecimal` only accepts a `Boolean` or a `String`, neither of which can be expressed as an uncertainty. But I see that the test cases you created cause the CQL-to-ELM translator to automatically insert a `ToDecimal` call, passing in an uncertainty -- so it seems the translator may invoke things technically not in the spec!

I guess we probably should support something like this for some of the other `ToXXX` functions, particularly those that _can_ take an argument that can be represented by an Uncertainty.  That would be:
```
ToBoolean(argument Decimal) Boolean
ToBoolean(argument Integer) Boolean
ToDate(argument DateTime) Date
ToDateTime(argument Date) DateTime
ToQuantity(argument Decimal) Quantity
ToQuantity(argument Integer) Quantity
```

I think I would also add `ToInteger`, because I think the translator maybe would use something like that to try to convert a Decimal to an Integer.

I left `toString` off the list because the spec does not define how to translate and uncertainty to a String and it also says that the string must be "round-trippable" back to its original type.

All the said, I think it's outside the scope of this task.  So I'd suggest we create another bug to track it.

cmoesel (approved):
This looks good!  I added a comment about a new issue we should log, but I think it is outside the scope of this PR!

PR #232, opened 2021-03-03
Record evaluated records on a patient-by-patient basis

cmoesel:
There was a bug that caused the Results.evaluatedRecords array to only contain the evaluated records of the most recently processed patient.

After discussing w/ the abacus team, we decided to switch this to record evaluated records patient by patient.  A new Results.patientEvaluatedRecords property was introduced, which is an object w/ patient ids as keys.

I don't think anyone is using the old Results.evaluatedRecords array, but just in case, I added a getter to provide backward-compatibility.

Surprisingly (and somewhat disturbingly) this did not break any tests.

**Submitter:**

**Reviewer:** mgramigna

Name:

### DIFF
--- a/examples/browser/cql4browsers.js
+++ b/examples/browser/cql4browsers.js
@@ -13510,6 +13510,8 @@ module.exports = {
 },{"../datatypes/datatypes":6,"../datatypes/exception":8,"../util/util":48}],42:[function(require,module,exports){
 "use strict";
 
+function _defineProperty(obj, key, value) { if (key in obj) { Object.defineProperty(obj, key, { value: value, enumerable: true, configurable: true, writable: true }); } else { obj[key] = value; } return obj; }
+
 function _classCallCheck(instance, Constructor) { if (!(instance instanceof Constructor)) { throw new TypeError("Cannot call a class as a function"); } }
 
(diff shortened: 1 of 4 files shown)


mgramigna:
Yeah.. it looks like the tests still pass because it makes assertions on the `evaluatedRecords` property of the context, which is pushed to during the actual Retrieve expression execution. Then we use that array to form the `patientEvaluatedRecords` object.

do you think the `patientEvaluatedRecords` should be covered by the unit tests rather than accessing the property directly off `ctx`?

cmoesel:
I think that ideally we'd have tests at both levels.  So I would not propose removing the assertions that are already in place. I thought about adding assertions to our `Executor` tests, but then I discovered we have no `Executor` tests. Oops.  And I didn't have time to create a whole new test suite -- so I figured it might be ok since putting up a PR w/ no tests for an update to a feature that already had no tests is basically a lateral move. ;-)

_EDIT: Hey!  I just found the executor tests!  They were not where I expected them to be!  So... I'll add a quick assertion or two so we at least have some coverage of the `patientEvaluatedRecords`._

mgramigna:
Nice! Once we get that test it, this looks good to me. I have tested locally and within fqm-execution and have confirmed that this will resolve the issue present in how we use `evaluatedRecords`

cmoesel:
OK, @mgramigna -- tests added, CI is passing.

mgramigna (approved):
Just checked everything again locally. Looks good to me.

Thanks for implementing this fix!

PR #235, opened 2021-03-18
ToQuantity: Fix Decimal/Integer Overloads & add Ratio Overload

### LINKED ISSUE #175: Support "ToQuantity" operator for Ratio type
See: https://cql.hl7.org/09-b-cqlreference.html#toquantity

cmoesel:
This PR contains fixes and enhancements to `ToQuantity`:
* Decimal and Integer overloads now properly set units to `1`
* Ratio overload is now supported (divides ratio numerator by denominator)

Fixes #175 

**Submitter:**

**Reviewer:**

Name:

### DIFF
--- a/examples/browser/cql4browsers.js
+++ b/examples/browser/cql4browsers.js
@@ -11771,6 +11771,7 @@ var _require3 = require('../datatypes/clinical'),
     Concept = _require3.Concept;
 
 var _require4 = require('../datatypes/quantity'),
+    Quantity = _require4.Quantity,
     parseQuantity = _require4.parseQuantity;
 
 var _require5 = require('../util/math'),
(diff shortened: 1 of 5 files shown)


cmoesel:
I was a bit perplexed how coverage went down, particularly in a file I did not modify.  But it is because we used to call `parseQuantity` for _all_ cases, which is clearly not the right thing to do.  One of those cases (for numbers) set the units to `''` in `quantity.js` -- but since that's not the behavior we want, we don't call that function that way anymore, so that line that sets units to `''` is no longer covered.  Honestly... I'm not worried about it -- but let me know if you are.

kjmahalingam (approved):
Looks great! I’m not concerned about that very slight decrease in code coverage because it it’s in line with what we’d expect here.

PR #239, opened 2021-03-31
bumping cql-to-elm to version 1.5.2. s

rdingwell:
As the title says this PR is for bumping the version of cql-to-elm to version 1.5.2. This is  to allow for functionality that is contained within CQL 1.5.x to be developed within the execution engine. 

Changes made to cql-to-elm 1.5.2 caused the functionality to load model files to no longer to work .  Contained in this PR is a fix to JavaScriptTestDataGenerator to deal with those changes and allow for the custom data model to be loaded for use in the tests. 

This also includes a regeneration of the test data based off of the 1.5.2 version of the library.  

This also includes an update to a dependency to take care of a yarn audit issue that presented while creating the PR. 

**Submitter:**
- N/A Tests are included and test edge cases

**Reviewer:**

Name:

### DIFF
--- a/examples/browser/cql4browsers.js
+++ b/examples/browser/cql4browsers.js
@@ -7516,7 +7516,12 @@ var Expand = /*#__PURE__*/function (_Expression14) {
       var _this$execArgs17 = this.execArgs(ctx),
           _this$execArgs18 = _slicedToArray(_this$execArgs17, 2),
           intervals = _this$execArgs18[0],
-          per = _this$execArgs18[1];
+          per = _this$execArgs18[1]; // CQL 1.5 introduced an overload to allow singular intervals; make it a list so we can use the same logic for either overload
+
+
+      if (!Array.isArray(intervals)) {
+        intervals = [intervals];
+      }
 
       var type = intervalListType(intervals);
 
--- a/examples/browser/cql4browsers.js
+++ b/examples/browser/cql4browsers.js
@@ -115,7 +115,7 @@ function _possibleConstructorReturn(self, call) { if (call && (_typeof(call) ===
 
 function _assertThisInitialized(self) { if (self === void 0) { throw new ReferenceError("this hasn't been initialised - super() hasn't been called"); } return self; }
 
-function _isNativeReflectConstruct() { if (typeof Reflect === "undefined" || !Reflect.construct) return false; if (Reflect.construct.sham) return false; if (typeof Proxy === "function") return true; try { Date.prototype.toString.call(Reflect.construct(Date, [], function () {})); return true; } catch (e) { return false; } }
+function _isNativeReflectConstruct() { if (typeof Reflect === "undefined" || !Reflect.construct) return false; if (Reflect.construct.sham) return false; if (typeof Proxy === "function") return true; try { Boolean.prototype.valueOf.call(Reflect.construct(Boolean, [], function () {})); return true; } catch (e) { return false; } }
 
 function _getPrototypeOf(o) { _getPrototypeOf = Object.setPrototypeOf ? Object.getPrototypeOf : function _getPrototypeOf(o) { return o.__proto__ || Object.getPrototypeOf(o); }; return _getPrototypeOf(o); }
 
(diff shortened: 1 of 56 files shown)


cmoesel:
Unfortunately, there are still issues w/ the CQL-to-ELM generation of tests; but this time it is with the spec-tests (which are new since you last worked on this project).  If you run `yarn build:all`, you'll see a bunch of errors like this near the end:
```
================================================================================
TRANSLATE /Users/cmoesel/dev/cqframework/cql-execution/test/generator/../spec-tests/cql/CqlErrorsAndMessagingOperatorsTest.cql
Translation failed due to errors:
Error:[2:1, 2:12] Could not resolve model info provider for model Simple, version null.
Error:[3:1, 3:15] Could not resolve context name Patient in model System.
```

This happens compiling the CQL [here](https://github.com/cqframework/cql-execution/tree/master/test/spec-tests/cql) (which is generated from the test cases the CQL spec provides).  The tests don't compile without specifying a model, so we specify the same `Simple` model we use in the other tests.  We run the translator by invoking the CLI interface [here](https://github.com/cqframework/cql-execution/blob/master/test/generator/build.gradle#L29-L33) -- and that's failing due to the change (actually, _bug_) in model loading.

I tried changing the CQL to use FHIR instead, but the translator still issues errors (different ones this time) and two tests fail.  It would be fairly annoying to have to write more Java code to get around this.  I wonder if we can get Bryn to make a quick fix and do a 1.5.3?

brynrhodes:
@cmoesel , what is the bug in model loading? Happy to fast track a fix for that, just need to know what to fix

cmoesel:
@brynrhodes -- I had no idea you were following the issues on this repository. Thanks!  So, here's the basic issue.  We used to be able to load arbitrary models like this:
```
cql-to-elm --input path/to/cql --model path/to/custom-modelinfo.xml
```
This doesn't seem to work anymore as we get errors like this (for CQL using a custom model called `Simple`):
> Error:[2:1, 2:12] Could not resolve model info provider for model Simple, version null.

That `--model` flag calls [CQLTranslator.loadModelInfo](https://github.com/cqframework/clinical_quality_language/blob/master/Src/java/cql-to-elm/src/main/java/org/cqframework/cql/cql2elm/CqlTranslator.java#L667-L673), which is defined like this:
```java
public static void loadModelInfo(File modelInfoXML) {
    final ModelInfo modelInfo = JAXB.unmarshal(modelInfoXML, ModelInfo.class);
    final VersionedIdentifier modelId = new VersionedIdentifier().withId(modelInfo.getName()).withVersion(modelInfo.getVersion());
    final ModelInfoProvider modelProvider = (VersionedIdentifier modelIdentifier) -> modelInfo;
    final ModelInfoLoader modelInfoLoader = new ModelInfoLoader();
    modelInfoLoader.registerModelInfoProvider(modelProvider);
}
```
Note that `modelInfoLoader` is just a locally scoped variable.  So the `modelProvider` gets registered on it, but since it is locally scoped, it goes away as soon as the method exits.  In version 1.5.1, the `modelProvider` was registered via a static method on `ModelInfoLoader` instead -- so it was a more "global" registration.

cmoesel:
Actually, @rdingwell, think I found a workaround we can use for now.  If I specify the QUICK 3.3.0 model instead, the translator seems to work fine (since QUICK is built in).  It doesn't matter to us what actual model we use since the spec tests don't actually invoke any model types.

Those two tests still fail, but those are actually because the translator is no longer converting `expand` arguments to lists:
image
This is actually valid now since [CQL 1.5 introduced an Interval<T> overload for Expand](http://build.fhir.org/ig/HL7/cql/09-b-cqlreference.html#expand) -- so it's a real bug to be fixed in `cql-execution`.  It should be easy.  I'll go ahead and do it and push it up to this branch.

@brynrhodes -- the bug I noted in my previous comment is still a real bug that I think should be fixed, but for now, we have a workaround so it's not as urgent.

cmoesel (approved):
This looks good from my perspective, but @rdingwell should review my last two commits before we merge.

rdingwell:
@cmoesel Looks good.  I have to get a bit more aquatinted with the all of the changes that have been made.

Issue #238, opened 2021-03-30
Null should be treated as empty list in Union


cmoesel:

cmoesel:
Fixed by #242

PR #242, opened 2021-04-02
Fixing issue related to issue #238

rdingwell:
This fixes the issue with performing unions between lists when one of the lists is null as identified in #238 

**Submitter:**
- [n/a] All dependent libraries are appropriately updated or have a corresponding PR related to this change

**Reviewer:**

Name:

### DIFF
--- a/src/elm/overloaded.js
+++ b/src/elm/overloaded.js
@@ -60,18 +60,24 @@ class Union extends Expression {
   exec(ctx) {
     const [a, b] = this.execArgs(ctx);
     if (a == null && b == null) {
-      return null;
+      return this.listTypeArgs() ? [] : null;
     }
     if (a == null || b == null) {
       const notNull = a || b;
       if (typeIsArray(notNull)) {
         return notNull;
       } else {
-        return [];
+        return null;
       }
     }
     const lib = typeIsArray(a) ? LIST : IVL;
     return lib.doUnion(a, b);
+  }
+
+  listTypeArgs() {
+    return this.args.every(arg => {
+      return arg.asTypeSpecifier != null && arg.asTypeSpecifier.type === 'ListTypeSpecifier';
+    });
   }
 }
 
--- a/src/elm/overloaded.js
+++ b/src/elm/overloaded.js
@@ -59,8 +59,16 @@ class Union extends Expression {
 
   exec(ctx) {
     const [a, b] = this.execArgs(ctx);
-    if (a == null || b == null) {
+    if (a == null && b == null) {
       return null;
+    }
+    if (a == null || b == null) {
+      const notNull = a || b;
+      if (typeIsArray(notNull)) {
+        return notNull;
+      } else {
+        return null;
+      }
     }
     const lib = typeIsArray(a) ? LIST : IVL;
     return lib.doUnion(a, b);
--- a/src/elm/overloaded.js
+++ b/src/elm/overloaded.js
@@ -75,7 +75,7 @@ class Union extends Expression {
   }
 
   listTypeArgs() {
-    return this.args.every(arg => {
+    return this.args.some(arg => {
       return arg.asTypeSpecifier != null && arg.asTypeSpecifier.type === 'ListTypeSpecifier';
     });
   }
--- a/src/elm/overloaded.js
+++ b/src/elm/overloaded.js
@@ -67,7 +67,7 @@ class Union extends Expression {
       if (typeIsArray(notNull)) {
         return notNull;
       } else {
-        return null;
+        return [];
       }
     }
     const lib = typeIsArray(a) ? LIST : IVL;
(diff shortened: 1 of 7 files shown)


src/elm/overloaded.js
```diff
@@ -59,9 +59,17 @@ class Union extends Expression {
 
   exec(ctx) {
     const [a, b] = this.execArgs(ctx);
-    if (a == null || b == null) {
+    if (a == null && b == null) {
```
cmoesel, line 62: Well... this is kind of a tricky thing, but technically the list union says:
> If either argument is null, it is considered an empty list for the purposes of evaluating the union.

Which means that `(null as List<String>) union (null as List<String>)` should actually be `List<String>{}`, not `null` (since empty list union empty list is empty list).

But it is difficult to determine the type of null, isn't it?  That said, there are two things we _could_ do if we wanted:
* Check to see if the ELM argument has a `resultTypeSpecifier` element or `resultTypeName` attribute.  These will only be present if the ELM was translated with the `result-types` option set.
* Check to see if the ELM argument is an `As` operation casting the argument to a list.

Is it worth going through this trouble?  I'm not sure.  What do you think?

cmoesel (commented):
This generally looks good, but I had one point for some discussion.

src/elm/overloaded.js
rdingwell: I had the same thoughts and erred on the not having the type information in the function call that I could not tell if it was null lists or null intervals.   Do we have methods for determining the result or argument types?  If not I can look into what it would take to add them.   

May need to kick this back up to the spec as well.  I can see returning an empty list if there are types given somewhere.  But if the call is actually `null union null ` there isn't any type information available.  In that case is List assumed and the return value is {} or are they treated as intervals and null is returned?

rdingwell: actually it looks like `null union null` gets converted to as List in cql-to-elm so looking at the type of the args might be possible

rdingwell: Playing around with different combinations of expressions it looks like many things are handled by the translator, converting things to lists before hand by either translating them inline to a list or calling ToList on the args.  It maybe that the only chance both args are null is when there is something like null as List<String>.   So instead of returning null it may be that returning an empty list is what is expected.  What do you think?

cmoesel: I'm not sure I understand your last comment.  I suspect you can do `null as Interval<Integer> union null as Interval<Integer>`, right?  Of course using literal `null` in CQL anywhere is already fairly unrealistic.  The much more common case will be null intervals and lists as the result of accessing properties w/ no value.  E.g.
```
define ForcedExample:
  for Observation O1, Observation O2
  return ToInterval(O1.effectivePeriod) union ToInterval(O2.effectivePeriod)
```
(BTW -- I didn't check that; might not compile exactly as is).

So... what are you proposing?  Can you rephrase?

If we don't know if the result should be an interval or a list, then returning `null` is _probably_ safer than returning `{}` -- because code expecting to get an `Interval` may not handle an empty list gracefully, but a `null` _should_ always be handled gracefully (since we've come to expect nulls).

rdingwell: I was so focused on Lists that I didn't consider the ToInterval scenarios.  I'll see what can be gleaned from the args and go from there.

cmoesel: I don't think it's feasible to have a perfect solution that works 100% of the time, but if there is a reasonable 80% solution, it's probably worth considering.  But I don't want you to invest another 20 hours in this.

rdingwell: Ok, I think the last commit should deal work.  When both arguments are null it checks to see if they have been cast to lists and if so it returns an empty list. Otherwise if both args are null it returns null.

src/elm/overloaded.js
```diff
@@ -60,19 +60,25 @@ class Union extends Expression {
...
       }
     }
     const lib = typeIsArray(a) ? LIST : IVL;
     return lib.doUnion(a, b);
   }
+
+  listTypeArgs() {
+    return this.args.every(arg => {
+      return arg.asTypeSpecifier != null && arg.asTypeSpecifier.type === 'ListTypeSpecifier';
+    });
```
cmoesel, line 80: Since the allowed arguments to `union` are either `List, List` or `Interval, Interval`, I don't think it's necessary to check `every` arg.  I think checking for `some` arg would be sufficient and would actually be better, since cases where _both_ args are cast are probably contrived.

cmoesel (changes requested):
Checking for casts still is not _perfect_ (as there will be invocations where neither argument is cast), but it's definitely closer, and probably "close enough" for now.  I've made one suggestion that I think will make this useful across at least a slightly broader set of use cases.

cmoesel (approved):
Looks good and thank you!

Issue #204, opened 2021-01-08
Engine cannot tell if whole numbers are Integers or Decimals


cmoesel:
Since we use standard JavaScript `number`, we lose the distinction of integer vs decimal.  For example, the Decimal `2.0` just becomes the number `2` in the JS runtime -- and we can't tell if it is supposed to be a decimal or integer.  This affects operations like `predecessor` and `successor`, as well as operations that check for overflows and underflows (since integers and decimals have different min/max).

This will likely require us to integrate a more sophisticated representation of numbers and may be a big deal.

PR #245, opened 2021-06-09
Implement Precision operator for DateTime/Date/Time

### LINKED ISSUE #160: Support "Precision" operator
See: https://cql.hl7.org/09-b-cqlreference.html#precision

kjmahalingam:

| **Test** | **Error** |
| --- | --- |
| `CqlArithmeticFunctionsTest.Precision.PrecisionYear` | Precision not implemented |
| `CqlArithmeticFunctionsTest.Precision.PrecisionDateTimeMilliseconds` | Precision not implemented |
| `CqlArithmeticFunctionsTest.Precision.PrecisionTimeMinutes` | Precision not implemented |
| `CqlArithmeticFunctionsTest.Precision.PrecisionTimeMilliseconds` | Precision not implemented |

Does *not* fix `PrecisionDecimal` since this requires additional work with integers and decimals as in #204 

Fixes #160 

**Submitter:**

**Reviewer:**

Name:

### DIFF
--- a/src/datatypes/datetime.js
+++ b/src/datatypes/datetime.js
@@ -342,6 +342,25 @@ class DateTime {
     return result;
   }
 
+  getPrecisionValue() {
+    const precisionValueMap = new Map();
+    if (this.isTime()) {
+      precisionValueMap.set(DateTime.Unit.HOUR, 2);
+      precisionValueMap.set(DateTime.Unit.MINUTE, 4);
+      precisionValueMap.set(DateTime.Unit.SECOND, 6);
+      precisionValueMap.set(DateTime.Unit.MILLISECOND, 9);
+    } else {
+      precisionValueMap.set(DateTime.Unit.YEAR, 4);
+      precisionValueMap.set(DateTime.Unit.MONTH, 6);
+      precisionValueMap.set(DateTime.Unit.DAY, 8);
+      precisionValueMap.set(DateTime.Unit.HOUR, 10);
+      precisionValueMap.set(DateTime.Unit.MINUTE, 12);
+      precisionValueMap.set(DateTime.Unit.SECOND, 14);
+      precisionValueMap.set(DateTime.Unit.MILLISECOND, 17);
+    }
+    return precisionValueMap.get(this.getPrecision());
+  }
+
   toLuxonDateTime() {
     const offsetMins =
       this.timezoneOffset != null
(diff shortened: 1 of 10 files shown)


src/datatypes/datetime.js
```diff
@@ -342,6 +342,25 @@ class DateTime {
...
+      precisionValueMap.set(DateTime.Unit.YEAR, 4);
+      precisionValueMap.set(DateTime.Unit.MONTH, 6);
+      precisionValueMap.set(DateTime.Unit.DAY, 8);
+      precisionValueMap.set(DateTime.Unit.HOUR, 10);
+      precisionValueMap.set(DateTime.Unit.MINUTE, 12);
+      precisionValueMap.set(DateTime.Unit.SECOND, 14);
+      precisionValueMap.set(DateTime.Unit.MILLISECOND, 17);
+    }
+    return precisionValueMap.get(this.getPrecision());
+  }
```
cmoesel, line 362: I think it would be more performant to initialize these maps as constants in the file -- and then just do something like:
```js
const pvMap = this.isTime() ? TIME_PRECISION_VALUE_MAP : DATETIME_PRECISION_VALUE_MAP;
return pvMap.get(this.getPrecision());
```

src/datatypes/datetime.js
```diff
@@ -623,6 +642,14 @@ class Date {
     return result;
   }
 
+  getPrecisionValue() {
+    const precisionValueMap = new Map();
+    precisionValueMap.set(DateTime.Unit.YEAR, 4);
+    precisionValueMap.set(DateTime.Unit.MONTH, 6);
+    precisionValueMap.set(DateTime.Unit.DAY, 8);
+    return precisionValueMap.get(this.getPrecision());
+  }
```
cmoesel, line 651: Same here -- I'd recommend initializing a constant map, or you could probably use the same map as DateTime uses.

src/elm/arithmetic.js
```diff
@@ -440,7 +440,7 @@ class Successor extends Expression {
   exec(ctx) {
     const arg = this.execArgs(ctx);
     if (arg == null) {
-      null;
+      return null;
```
cmoesel, line 443: Whoops!  Good catch!

cmoesel (changes requested):
Implementation looks good, but I'd prefer an approach that doesn't create a new Map every time the function is invoked.

src/datatypes/datetime.js
kjmahalingam: I actually did something like this initially, and then realized that for some reason the `DateTime` namespace was out of scope of where I was defining them at the top of the file. Maybe I can define the maps within the classes themselves, and that should take care of the issue. I'll give it a shot!

src/datatypes/datetime.js
kjmahalingam: Will do the same here!

src/datatypes/datetime.js
kjmahalingam: Turns out the issue was an ordering thing; I put them at the top of the file before `DateTime.Unit` was available. I just put them with the other similar constants and now it's all working.

PR #243, opened 2021-05-17
Feature/include codesystems

poolis:
I believe the offending CQL is this...

```
...
include Terminology version '1.0.0' called Terminology
...
code "Institutional": 'institutional' from Terminology."claim-type" display 'Institutional'
code "Professional": 'professional' from Terminology."claim-type" display 'Professional'
...
define "Institutional or Professional Claims":
    [Claim: "Institutional"]
        union
        [Claim: "Professional"]
...
```

When the `Library` object is created, the `codesystems` in the included libraries are not in the `codesystems` of the `Library` object, despite the included libraries being in the `includes` parameter.  This causes a lookup failure and processing to error out. With this change, execution seems to continue as expected.

PR #249, opened 2021-07-12
Add boolean logic to repository resolve that handles absent version

mgramigna:
Per the [ELM spec](https://cql.hl7.org/04-logicalspecification.html#includedef), `version` is 0..1 for an `IncludeDef`. The `resolve` method of the `Repository` class currently checks for a match with both ID and version.

This causes an error during `resolve` where `include`s without a version won't be resolved properly. This PR adds a fallback case to just check by `id` when version is not specified during a `resolve` call.

I've also added some simple tests for the `Repository` class that test resolves with a simple `include` definition.

---

**Submitter:**

**Reviewer:**

Name:


test/runtime/repository.test.js
```diff
@@ -0,0 +1,42 @@
...
+
+  it('should fail when version is incorrect', () => {
+    simpleELMJson.library.includes.def[0].version = 'fake-version';
+
+    const repository = new Repository([simpleELMJson, simpleELMDependency]);
+
+    // When resolution can't happen, an error is thrown
+    should(() =>
+      repository.resolve(simpleIncludeIdentifier.id, simpleIncludeIdentifier.version)
+    ).throw();
```
cmoesel, line 40: This confused me at first, as I couldn't figure out why it was throwing.  Thanks for walking me through it.  As discussed, it's throwing because PR #243 introduced a block of code that doesn't null-check the included library before trying to access its code systems.  So the behavior of throwing when constructing the main lib is actually unintentional.

I think that the overall approach for how included libraries' codesystems are handled needs to be revisited, but that is not your concern.  Still, would you mind fixing [this](https://github.com/mgramigna/cql-execution/blob/a559aea61afcedcd99496daf5092c5ea73a1a9fb/src/elm/library.js#L66) so it null checks the library before accessing its codesystems property?  That way we will no longer throw there.  And then update this test to check that the dependency lib is undefined from the main lib (since I guess that is what is expected)?

It also might be good to do direct `resolve` on the dependency lib and assert it is undefined (instead of only going through the main lib).

cmoesel (changes requested):
Thanks, @mgramigna!  As discussed, you actually exposed a different issue in the code, but one that I think should probably be addressed (in a minimal fashion) as part of this PR.  See the individual comment for more details.

test/runtime/repository.test.js
mgramigna: Good call. I simplified the approach a bit to use one fixture and test resolve that way.

I also added another guard in the library constructor line that you linked.

mgramigna:
after an onslaught of broken commits, this branch should finally be good now

cmoesel (approved):
Looks good!  Thanks, @mgramigna!

PR #251, opened 2021-08-20
Support resolving with path.

bkaney:
When the IG Publisher (using binary adjunct support) converts CQL to ELM it creates a system in the library identifier using the canonical base, and creates a full URI for the path in includes.

Here is the identifiers having `system` specified as a URI:

```json
{
  "library" : {
     "annotation" : [ {
        "translatorOptions" : "EnableAnnotations,EnableLocators,DisableListDemotion,DisableListPromotion",
        "type" : "CqlToElmInfo"
     } ],
     "identifier" : {
        "id" : "HypertensionCaseFeatureLogic",
        "system" : "http://acme.com/canonical-base",
        "version" : "0.1.0"
     },
     …
```

And `includes` has a proper URI built for the `path` portion of include, as seen below:

```json
  "includes": {
    "def": [
      {
        "localId": "2",
        "locator": "5:1-5:54",
        "localIdentifier": "FHIRHelpers",
        "path": "http://hl7.org/fhir/FHIRHelpers",
        "version": "4.0.1",
      ...
```

This patch allows it to work when system is used and the resolve path is a full path.

**Submitter:**
- [ ] (N/A) All dependent libraries are appropriately updated or have a corresponding PR related to this change

**Reviewer:**

Name:


cmoesel:
Thanks for the PR, @bkaney!  We really appreciate it! I have a few comments/questions regarding the PR, which I'll note below.

I see that you added `test/elm/library/data-full-paths.js` directly w/ no corresponding CQL file.  The way our ELM-based tests are designed is that we have a source CQL file (w/ special annotations) that sets up the tested CQL expressions (see [test/elm/library/data.cql](https://github.com/Vermonster/cql-execution/blob/support-path-in-resolve/test/elm/library/data.cql).  Then you can run `yarn run build:test-data` and it will automatically generate the corresponding `data.js` that can be required and used in the test JS.  This makes it so we don't have to hand-modify the ELM JSON and also ensures the ELM JSON reflects the real output of whatever CQL-to-ELM translator we're currently targeting.  Would it be possible to update your test data to use that approach?  I don't recall exactly how you signal that the system should be used in the ELM, however, so let me know if this is not feasible.

In addition, it looks like lint or prettier is failing.  Could you take a look at that?

And last, it was failing due to a `yarn audit`.  I just merged a dependabot PR to master, so if you rebase or merge master in, that might resolve that issue.

Thanks again!  Let me know if you have any questions or concerns!

cmoesel:
I should have mentioned this yesterday, but we have some (brief) documentation on writing tests here: https://github.com/cqframework/cql-execution#to-develop-tests

bkaney:
Hey @cmoesel  -- I _think_ I see the problem, which we can use your standard testing setup to reproduce. The cql-to-elm version in IG Publisher is 1.5.0 (see https://github.com/HL7/fhir-ig-publisher/blob/master/org.hl7.fhir.publisher.core/pom.xml#L38-L42) and you have 1.5.2 (https://github.com/cqframework/cql-execution/blob/master/test/generator/build.gradle#L13)

Incidentally 1.5.3 is the most recent now - https://oss.sonatype.org/#nexus-search;quick~cqframework

We are using the IG Publisher toolchain, so it would be nice to support the output of whatever it creates...

I can't simply change it back to 1.5.0 (in `build.gradle`) and fails with a type error, so this will have to be looked into more.

cmoesel:
@bkaney -- are you saying that the 1.5.2 CQL-to-ELM does not create the system namespaces in the ELM (but the 1.5.0 did)?  If so, that's probably something to talk to Bryn about.  If it's possible to get the 1.5.2 version to populate the systems though, I think that would be preferable.

All that said, if you specifically need to test something that only exists in a particular version/branch of the CQL-to-ELM translator, then that would be a good justification for just checking in the ELM JSON directly.  If that is indeed the case, then I'd just suggest adding a comment in the test somewhere in order to call this out so others understand what is going on.

cmoesel (approved):
This looks good!  Thanks, @bkaney!

src/runtime/repository.js
```diff
@@ -10,13 +10,14 @@ class Repository {
     for (const lib of this.libraries) {
       if (lib.library && lib.library.identifier) {
         const id = lib.library.identifier;
+        const path = (id.system ? id.system + '/' : '') + id.id;
 
         // Check version immediately if specified, fallback to ID otherwise
         if (version) {
-          if (id.id === library && id.version === version) {
+          if (path === library && id.version === version) {
```
mgramigna, line 17:
```suggestion
          if ((path === library || id.id === library) && id.version === version) {
```

See root-level comment, we should be falling back to the original check if the canonical check failed

src/runtime/repository.js
```diff
@@ -10,13 +10,14 @@ class Repository {
...
             return new Library(lib, this);
           }
-        } else if (id.id === library) {
+        } else if (path === library) {
```
mgramigna, line 20:
```suggestion
        } else if (path === library || id.id === library) {
```

See root-level comment, we should be falling back to the original check if the canonical check failed

mgramigna (changes requested):
Thanks for the pull request! These changes uncovered an edge case in the ELM that we use for eCQMs.

There can be cases in ELM where a library's identifier is defined using a system, but the `include` def from a different library does not reference it by canonical. E.g.:

```
// Hospice library identifier
"identifier": {
      "id": "Hospice",
      "system": "http://fhir.org/guides/dbcg/connectathon",
      "version": "2.0.000"
    }
```

```
// other library including Hospice
"includes": {
      "def": [
        {
          "localId": "3",
          "locator": "17:1-17:48",
          "localIdentifier": "Hospice",
          "path": "Hospice",
          "version": "2.0.000"
        }
      ]
}
```

In the above case, this code will now check if `'Hospice' === 'http://fhir.org/guides/dbcg/connectathon/Hospice`. I've added inline suggestions to, in the case of the canonical check failing, fall back to a check for just `id.id === library`, since we still want to support cases like the above example.

It might help to add a test case for this as well. Perhaps we could add to the existing test cases you've written to manually change the `path` of one of the include statements in the example libraries such that it doesn't reference by canonical, and ensure that the library can still resolve properly?

Let me or @cmoesel know if you have any questions. Sorry for the long rambling. (Side note: thanks for `@types/fhir`, we use it all the time)

bkaney:
Hi @mgramigna @cmoesel  --

Reading through the schema (https://cql.hl7.org/elm/schema/library.xsd), I made an initial assumption that concatenating the Identifier `system` and `id` would be how to construct the path to the library. However, Identifier `system` and `id` according to the schema are both`xs:string`. Whereas the `path` in IncludesDef is type `xs:anyURI`. 

It seems a bit odd to combine two strings to form a URI. Can we confirm this is correct assumption, this doesn't seem quite right? If it is correct assumption, I can make a path to cover all combinations (including with and without slashes).

cmoesel:
I was not involved in most of the thinking/implementation of how namespaces fit into the picture.  @brynrhodes -- could you speak to how one should expect an `IncludeDef.path` to relate to a `Library.identifier` (with and without the `system` specified)?  Is it safe to assume that the `system` is a URL?  Could it be a `urn:oid:xxx` or even just some arbitrary string?  And if so, how do you combine it w/ the id to make a fully qualified path?

brynrhodes:
That's correct, the `system` and `version` elements are combined to create the library uri. From the translator's perspective, the contents of the `system` element (and the corresponding contents of the `uri` element of the `NamespaceInfo` are opaque. It always and only deals with them as strings and uses the index of the last path separator to split a library uri into namespace uri and library name.

Note that the ELM Library URI is not the same as the FHIR Library resource URL, but there is a bi-directional transform:

    ELM Library URI: <namespace-uri>/<library-name>
    FHIR Library Resource URL: <namespace-uri>/Library/<library-name>

bkaney:
@cmoesel @mgramigna - please re-review. Thanks!

mgramigna (approved):

bkaney:
I would like to rework this. Please don’t merge yet. Since the spec is a string I don’t like looking for a /, as it may just be in the name of a library.

bkaney:
@mgramigna  @cmoesel  - I like this implementation better. It does not rely on heuristics of the path string that is passed in. Could you give this a review, I would be comfortable merging this in. Thanks!

mgramigna (approved):
Looks good to me, and works with our test cases

cmoesel (approved):
Yes, looks good!  Thanks for working through this with us!  We appreciate the contribution!

PR #263, opened 2022-02-10
Time values should be unspecified when converting Date using ToDateTime

dczulada:
https://cql.hl7.org/04-logicalspecification.html#todatetime
For Date values, the result is a DateTime with the time components unspecified, except the timezone offset, which is set to the timezone offset of the evaluation request timestamp.

Previously the time components are being set to 0.

**Submitter:**

**Reviewer:**

Name:


p9g:
Related to errata in CQL spec https://jira.hl7.org/browse/FHIR-35869

src/datatypes/datetime.ts
```diff
@@ -1179,7 +1179,7 @@ export class Date extends AbstractDate {
     // except for the timezone offset, which will be set to the timezone offset of the evaluation
     // request timestamp. (this last part is acheived by just not passing in timezone offset)
     if (this.year != null && this.month != null && this.day != null) {
-      return new DateTime(this.year, this.month, this.day, 0, 0, 0, 0);
+      return new DateTime(this.year, this.month, this.day, null, null, null, null);
```
cmoesel, line 1182: Line 1178 above (which GitHub will not let me directly comment on) says:
```ts
// from the spec: the result will be a DateTime with the time components set to zero,
```

That comment is now out of sync w/ the implementation.  Can you update it to reflect the correct behavior? It should read like this now:
```ts
// from the spec: the result will be a DateTime with the time components unspecified,
```

Also, while you're at it, could you fix the typo on line 1180 (`acheived`)?

test/elm/comparison/comparison-test.ts
```diff
@@ -71,20 +71,20 @@ describe('Equal', () => {
     this.impossiblyEqualDateTimes.exec(this.ctx).should.be.false();
   });
 
-  it('should be null for Date and DateTime equality with same year, month, hour', function () {
-    should(this.dateAndDateTimeNull.exec(this.ctx)).be.null();
+  it('should be true for Date and DateTime equality with same year, month, hour', function () {
+    this.dateAndDateTimeNull.exec(this.ctx).should.be.true();
```
cmoesel, line 75: The name of the CQL expression (and variable here) is now out of sync w/ the expectation.  Now that this returns `true`, the CQL expression should be `DateAndDateTimeTrue` (and corresponding variable here should be `dateAndDateTimeTrue`).

All of the test expressions and variables need to be checked and fixed to ensure their name still matches the expectation. (So I won't comment on anymore like this).

cmoesel (changes requested):
Thanks for this fix, @dczulada!  Everything seems to work great (and as expected).  I just found a few places where a comment and/or test expression or variable name needs to be changed.

cmoesel (approved):
Looks good to me!  Thanks, @dczulada!

mgramigna (approved):
Looks good and passed all my tests!

PR #265, opened 2022-02-24
Add passing of retrieve to findRecords implementations

mgramigna:
This PR expands the API for `findRecords` to include the entire ELM Retrieve expression instance as an optional second argument that can be utilized in implementations of `findRecords` (e.g. in `cql-exec-fhir`).

**Code Changes**:

* Modified the `exec` function of `Retrieve` class to pass along `this` to the `findRecords` call
* Bubble up the `retrieve` argument to all calls of `findRecords`
* Modify the `PatientObject` interface to specify the optional `retrieve` argument
  * Note: this does not break backwards compatibility, as TypeScript will allow this argument to be omitted from declarations of `findRecords` when not used. The `Patient` class in `cql-patient` is one example of this
* Added a unit tests that ensures that when a Retrieve is executed, the expression is passed along to the `findRecords` call of the Patient object used.

**Submitter:**

**Reviewer:**

Name:


cmoesel:
@mgramigna -- I think that, in my head, I was expecting to pass the `Retrieve` ELM JSON object, _not_ an instance of the `Retrieve` TypeScript class.  That said, I had forgotten that some of the values in the ELM JSON might be expressions that need to be evaluated -- so, in hindsight, passing just the JSON would put a lot of burden on implementers.

Since we're passing the class instance, I guess we might expect implementers to call `retrieve.codes.execute(ctx)` and `retrieve.dateRange.execute(ctx)` -- but the problem there is that we don't pass along the `ctx`, so they can't make those calls correctly.  It also requires a bit deeper knowledge of our inner workings than I would prefer.  _(Sigh)_.

Clearly I did not think this through quite enough before recommending it.

It kind of feels like whatever we pass to the PatientSource should already have its important properties resolved to concrete values (not expressions) -- which might mean two different `Retrieve` types: one w/ properties whose values are expressions (to be lazy-executed upon execution), the other w/ properties whose values are the results of said expressions.  The former is the instance that is _executed_, but the latter is the instance that is _passed_ to the PatientSource.  Does that make sense?  Is it feasible?  Or overly complex?  What do you think?

(Tagging @hossenlopp for his opinion as well).

mgramigna:
@cmoesel -- Just pushed up a new commit that _hopefully_ addresses the above.

It introduces a new interface that tracks the resolved information from the retrieve class, and reworks said class to do that computation before the findRecords call.

I also went ahead and updated some of the TypeScript types away from `any` since we know more about them now!

src/elm/external.ts
```diff
@@ -22,21 +25,49 @@ export class Retrieve extends Expression {
...
       }
-    }
-    if (codes) {
-      records = records.filter((r: any) => this.recordMatchesCodesOrVS(r, codes));
+
+      retrieveDetails.codes = resolvedCodes;
     }
     // TODO: Added @dateProperty check due to previous fix in cql4browsers in cql_qdm_patient_api hash: ddbc57
+    let range: Interval | undefined;
     if (this.dateRange && this.dateProperty) {
```
cmoesel, line 45: We checked for both `dateRange` and `dateProperty` in our code because we really needed both for it to be useful at all.  BUT... I think that for the purposes of passing this through to the provider, we shouldn't make such assumptions.

In other words, I'd suggest removing `&& this.dateProperty` from the condition.

src/elm/external.ts
```diff
@@ -22,21 +25,49 @@ export class Retrieve extends Expression {
...
-      const range = this.dateRange.execute(ctx);
-      records = records.filter((r: any) => range.includes(r.getDateOrInterval(this.dateProperty)));
+      range = this.dateRange.execute(ctx);
+
+      retrieveDetails.dateRange = range;
+      retrieveDetails.dateProperty = this.dateProperty;
```
cmoesel, line 49: And as noted above, I think we should avoid making assumptions about what properties are or are not useful in combination with other properties.  So I think that if `dateProperty` exists, even with no `dateRange`, we should go ahead and pass it through.  I can't imagine how it would be used, but there's no reason for us to suppress information that is in the ELM.

In other words, I suggest moving the assignment of `retrieveDetails.dateProperty` outside of this block.

src/elm/external.ts
```diff
@@ -22,21 +25,49 @@ export class Retrieve extends Expression {
...
+    }
+
+    if (this.templateId) {
+      retrieveDetails.templateId = this.templateId;
+    }
+
+    if (resolvedCodes && this.codeProperty) {
```
cmoesel, line 56: And to continue on the theme, I don't think we need to guard this assignment with the check on `resolvedCodes`.  In fact, we might not need to guard it at all, because if `this.codeProperty` is not defined, there's no harm is assigning it to `retrieveDetails.codeProperty` anyway, right?

src/elm/external.ts
```diff
@@ -22,21 +25,49 @@ export class Retrieve extends Expression {
...
+      retrieveDetails.codeProperty = this.codeProperty;
+    }
+
+    let records = ctx.findRecords(
+      this.templateId != null ? this.templateId : this.datatype,
+      retrieveDetails
+    );
+
+    if (resolvedCodes) {
```
cmoesel, line 65: Not a big deal, but... if you used `retrieveDetails.codes` here (and in the line below), then you wouldn't need `resolvedCodes` in this scope at all and you could just scope it inside the block on line 35 (as a `const`).

src/elm/external.ts
```diff
@@ -22,21 +25,49 @@ export class Retrieve extends Expression {
...
+      records = records.filter((r: any) => this.recordMatchesCodesOrVS(r, resolvedCodes));
+    }
+
+    if (range && this.dateProperty) {
```
cmoesel, line 69: Similarly to `resolvedCodes` above, if you used `retrieveDetails.range` then you could take `range` out of this scope and just keep it in the smaller scope as a `const`.

src/elm/external.ts
```diff
@@ -22,21 +25,49 @@ export class Retrieve extends Expression {
...
+      records = records.filter((r: any) => range?.includes(r.getDateOrInterval(this.dateProperty)));
```
cmoesel, line 70: Since you did a null-check before coming into this block, I don't think the `?` guard is necessary.

test/elm/external/external-test.ts
```diff
@@ -15,7 +15,11 @@ describe('Retrieve', () => {
     const findRecordsSpy = sinon.spy(this.ctx, 'findRecords');
     this.conditions.exec(this.ctx);
 
-    findRecordsSpy.should.be.calledWithExactly(this.conditions.datatype, this.conditions);
+    const expectedRetrieveDetail: RetrieveDetails = {
+      datatype: '{https://github.com/cqframework/cql-execution/simple}Condition'
+    };
```
cmoesel, line 20: This is the simplest of retrieves.  If it's not too much trouble, it would be nice to have a test that verifies that the `RetrieveDetails.code` property is correctly set to a calculated value (either a `Code[]` or `ValueSet`). Extra bonus points for a test that does the same for `RetrieveDetails.dateRange`

cmoesel (changes requested):
Oops.  I might have made a lot of comments by mistake. ;-)

src/elm/external.ts
mgramigna: You would think!!! Check this out:

Before the scope of the `if`, range is `Interval | undefined`. Logical.

image

Within the scope of the `if`, it's STILL `Interval | undefined`

image


BUT within the scope of the `if` and outside the anonymous function within `filter`, it's just `Interval`:

image


Legit no idea

src/elm/external.ts
mgramigna: I guess the only open question with this (and dateProperty), is if they're undefined, do we want them to be omitted entirely? Or show up on the object with a value of `undefined`?

cmoesel: From a practical point of view, it probably doesn't matter.  But from my (sometimes perfectionist) point of view, I prefer they be omitted entirely.

But... do I prefer it strong enough to require an `if` block (3 lines of code) vs a single line of code?  Hmmm...  Tough question.

I will be happy either way.  Follow your heart, @mgramigna.

mgramigna:
@cmoesel I think I got everything!

src/elm/external.ts
```diff
@@ -22,21 +25,48 @@ export class Retrieve extends Expression {
...
     }
-    if (codes) {
-      records = records.filter((r: any) => this.recordMatchesCodesOrVS(r, codes));
-    }
+
     // TODO: Added @dateProperty check due to previous fix in cql4browsers in cql_qdm_patient_api hash: ddbc57
-    if (this.dateRange && this.dateProperty) {
-      const range = this.dateRange.execute(ctx);
-      records = records.filter((r: any) => range.includes(r.getDateOrInterval(this.dateProperty)));
+    if (this.dateRange) {
```
cmoesel, line 47: Oops.  I just read the TODO on line 46 for the first time.  I told you to remove the `dateProperty` check, but the TODO suggests it should be there.  But I think that's because of it's (prior) use in the filter.  You still check for it before filtering below, so I think it's probably OK that we don't check for it here.

I just went through the history and tracked this back to June of 2017 (courtesy of JB).  ;-)

Anyway, I still don't think we need the dateProperty check here, do you?  If not, then I suggest removing the confusing TODO.

src/elm/external.ts
```diff
@@ -22,21 +25,48 @@ export class Retrieve extends Expression {
...
+      const range: Interval = this.dateRange.execute(ctx);
+
+      retrieveDetails.dateRange = range;
```
cmoesel, line 50: I didn't check to make sure tsc is happy, but I assume this can be simplified to a single statement...
```suggestion
      retrieveDetails.dateRange = this.dateRange.execute(ctx);
```

cmoesel (commented):
Nice tests!  You get the super ultra over-the-top bonus points!

src/elm/external.ts
cmoesel: Ah, sorry, it's not showing the comment in the context above (on the Conversation page).  Here it is:
```ts
// TODO: Added @dateProperty check due to previous fix in cql4browsers in cql_qdm_patient_api hash: ddbc57
```

mgramigna: agreed

cmoesel (approved):
Bingo!  Nicely done!  I think this sets us up nicely for some more advanced functionality in the PatientSource implementations.  Thanks, @mgramigna!

PR #267, opened 2022-03-14
Explore elm types

csenn:
This Pull Request starts the addition of ELM Typescript Types that can be used to make the ELM traversal logic a little more type safe.

**Submitter:**
- [ ✔] This pull request describes why these changes were made

**Reviewer:**

Name:

### DIFF
--- a/src/elm/arithmetic.ts
+++ b/src/elm/arithmetic.ts
@@ -11,7 +11,7 @@ import { Uncertainty } from '../datatypes/uncertainty';
 import { Context } from '../runtime/context';
 import { build } from './builder';
 import { DateTime } from '../datatypes/datetime';
-import ELM from '../types/elm'
+import ELM from '../types/elm';
 
 export class Add extends Expression {
   constructor(json: ELM.Add) {
@@ -390,14 +390,14 @@ export class MinValue extends Expression {
     '{urn:hl7-org:elm-types:r1}Time': MathUtil.MIN_TIME_VALUE
   };
 
-  valueType: keyof typeof MinValue.MIN_VALUES
+  valueType: keyof typeof MinValue.MIN_VALUES;
 
   constructor(json: ELM.MinValue) {
     super(json);
     if (json.valueType in MinValue.MIN_VALUES) {
-      this.valueType = json.valueType as keyof typeof MinValue.MIN_VALUES
+      this.valueType = json.valueType as keyof typeof MinValue.MIN_VALUES;
     } else {
-      throw new Error(`Not expecting MIN_VALUE: ${json.valueType})`)
+      throw new Error(`Not expecting MIN_VALUE: ${json.valueType})`);
     }
   }
 
@@ -430,9 +430,9 @@ export class MaxValue extends Expression {
   constructor(json: ELM.MaxValue) {
     super(json);
     if (json.valueType in MaxValue.MAX_VALUES) {
-      this.valueType = json.valueType as keyof typeof MaxValue.MAX_VALUES
+      this.valueType = json.valueType as keyof typeof MaxValue.MAX_VALUES;
     } else {
-      throw new Error(`Not expecting Max_VALUE: ${json.valueType})`)
+      throw new Error(`Not expecting Max_VALUE: ${json.valueType})`);
     }
   }
 
--- a/src/elm/arithmetic.ts
+++ b/src/elm/arithmetic.ts
@@ -11,9 +11,10 @@ import { Uncertainty } from '../datatypes/uncertainty';
 import { Context } from '../runtime/context';
 import { build } from './builder';
 import { DateTime } from '../datatypes/datetime';
+import ELM from '../types/elm'
 
 export class Add extends Expression {
-  constructor(json: any) {
+  constructor(json: ELM.Add) {
     super(json);
   }
 
@@ -56,7 +57,7 @@ export class Add extends Expression {
 }
 
 export class Subtract extends Expression {
-  constructor(json: any) {
+  constructor(json: ELM.Subtract) {
     super(json);
   }
 
@@ -94,7 +95,7 @@ export class Subtract extends Expression {
 }
 
 export class Multiply extends Expression {
-  constructor(json: any) {
+  constructor(json: ELM.Multiply) {
     super(json);
   }
 
@@ -132,7 +133,7 @@ export class Multiply extends Expression {
 }
 
 export class Divide extends Expression {
-  constructor(json: any) {
+  constructor(json: ELM.Divide) {
     super(json);
   }
 
@@ -172,7 +173,7 @@ export class Divide extends Expression {
 }
 
 export class TruncatedDivide extends Expression {
-  constructor(json: any) {
+  constructor(json: ELM.TruncatedDivide) {
     super(json);
   }
 
@@ -193,7 +194,7 @@ export class TruncatedDivide extends Expression {
 }
 
 export class Modulo extends Expression {
-  constructor(json: any) {
+  constructor(json: ELM.Modulo) {
     super(json);
   }
 
@@ -210,7 +211,7 @@ export class Modulo extends Expression {
 }
 
 export class Ceiling extends Expression {
-  constructor(json: any) {
+  constructor(json: ELM.Ceiling) {
     super(json);
   }
 
@@ -225,7 +226,7 @@ export class Ceiling extends Expression {
 }
 
 export class Floor extends Expression {
-  constructor(json: any) {
+  constructor(json: ELM.Floor) {
     super(json);
   }
 
@@ -240,7 +241,7 @@ export class Floor extends Expression {
 }
 
 export class Truncate extends Expression {
-  constructor(json: any) {
+  constructor(json: ELM.Truncate) {
     super(json);
   }
 
@@ -254,7 +255,7 @@ export class Truncate extends Expression {
   }
 }
 export class Abs extends Expression {
-  constructor(json: any) {
+  constructor(json: ELM.Abs) {
     super(json);
   }
 
@@ -271,7 +272,7 @@ export class Abs extends Expression {
 }
 
 export class Negate extends Expression {
-  constructor(json: any) {
+  constructor(json: ELM.Negate) {
     super(json);
   }
 
@@ -290,7 +291,7 @@ export class Negate extends Expression {
 export class Round extends Expression {
   precision: any;
 
-  constructor(json: any) {
+  constructor(json: ELM.Round) {
     super(json);
     this.precision = build(json.precision);
   }
@@ -307,7 +308,7 @@ export class Round extends Expression {
 }
 
 export class Ln extends Expression {
-  constructor(json: any) {
+  constructor(json: ELM.Ln) {
     super(json);
   }
 
@@ -324,7 +325,7 @@ export class Ln extends Expression {
 }
 
 export class Exp extends Expression {
-  constructor(json: any) {
+  constructor(json: ELM.Exp) {
     super(json);
   }
 
@@ -344,7 +345,7 @@ export class Exp extends Expression {
 }
 
 export class Log extends Expression {
-  constructor(json: any) {
+  constructor(json: ELM.Log) {
     super(json);
   }
 
@@ -361,7 +362,7 @@ export class Log extends Expression {
 }
 
 export class Power extends Expression {
-  constructor(json: any) {
+  constructor(json: ELM.Power) {
     super(json);
   }
 
@@ -381,7 +382,7 @@ export class Power extends Expression {
 }
 
 export class MinValue extends Expression {
-  static readonly MIN_VALUES = {
+  static MIN_VALUES = {
     '{urn:hl7-org:elm-types:r1}Integer': MathUtil.MIN_INT_VALUE,
     '{urn:hl7-org:elm-types:r1}Decimal': MathUtil.MIN_FLOAT_VALUE,
     '{urn:hl7-org:elm-types:r1}DateTime': MathUtil.MIN_DATETIME_VALUE,
@@ -389,11 +390,15 @@ export class MinValue extends Expression {
     '{urn:hl7-org:elm-types:r1}Time': MathUtil.MIN_TIME_VALUE
   };
 
-  valueType: keyof typeof MinValue.MIN_VALUES;
+  valueType: keyof typeof MinValue.MIN_VALUES
 
-  constructor(json: any) {
+  constructor(json: ELM.MinValue) {
     super(json);
-    this.valueType = json.valueType;
+    if (json.valueType in MinValue.MIN_VALUES) {
+      this.valueType = json.valueType as keyof typeof MinValue.MIN_VALUES
+    } else {
+      throw new Error(`Not expecting MIN_VALUE: ${json.valueType})`)
+    }
   }
 
   exec(ctx: Context) {
@@ -422,9 +427,13 @@ export class MaxValue extends Expression {
 
   valueType: keyof typeof MaxValue.MAX_VALUES;
 
-  constructor(json: any) {
+  constructor(json: ELM.MaxValue) {
     super(json);
-    this.valueType = json.valueType;
+    if (json.valueType in MaxValue.MAX_VALUES) {
+      this.valueType = json.valueType as keyof typeof MaxValue.MAX_VALUES
+    } else {
+      throw new Error(`Not expecting Max_VALUE: ${json.valueType})`)
+    }
   }
 
   exec(ctx: Context) {
@@ -443,7 +452,7 @@ export class MaxValue extends Expression {
 }
 
 export class Successor extends Expression {
-  constructor(json: any) {
+  constructor(json: ELM.Successor) {
     super(json);
   }
 
@@ -472,7 +481,7 @@ export class Successor extends Expression {
 }
 
 export class Predecessor extends Expression {
-  constructor(json: any) {
+  constructor(json: ELM.Predecessor) {
     super(json);
   }
 
--- a/src/elm/clinical.ts
+++ b/src/elm/clinical.ts
@@ -2,7 +2,7 @@ import { Expression } from './expression';
 import * as dt from '../datatypes/datatypes';
 import { Context } from '../runtime/context';
 import { build } from './builder';
-import ELM from '../types/elm'
+import ELM from '../types/elm';
 
 export class ValueSetDef extends Expression {
   name: string;
@@ -54,7 +54,7 @@ export class AnyInValueSet extends Expression {
     super(json);
     this.codes = build(json.codes);
     if (json.valueset?.type !== 'ValueSetRef') {
-      throw new Error('Expecting ValueSetRef Expression')
+      throw new Error('Expecting ValueSetRef Expression');
     }
     this.valueset = new ValueSetRef(json.valueset);
   }
@@ -79,7 +79,7 @@ export class InValueSet extends Expression {
     super(json);
     this.code = build(json.code);
     if (json.valueset?.type !== 'ValueSetRef') {
-      throw new Error('Expecting ValueSetRef Expression')
+      throw new Error('Expecting ValueSetRef Expression');
     }
     this.valueset = new ValueSetRef(json.valueset);
   }
@@ -196,7 +196,7 @@ export class ConceptDef extends Expression {
   display?: string;
 
   constructor(json: ELM.ConceptDef) {
-     // @ts-ignore
+    // @ts-ignore
     super(json);
     this.name = json.name;
     this.display = json.display;
@@ -215,7 +215,7 @@ export class ConceptDef extends Expression {
 export class ConceptRef extends Expression {
   name: string;
 
-  constructor(json:  ELM.ConceptRef) {
+  constructor(json: ELM.ConceptRef) {
     super(json);
     this.name = json.name || '';
   }
@@ -259,7 +259,7 @@ export class CalculateAge extends Expression {
   constructor(json: ELM.CalculateAge) {
     super(json);
     if (!json.precision) {
-      throw new Error('Precision is required')
+      throw new Error('Precision is required');
     }
     this.precision = json.precision;
   }
@@ -278,12 +278,12 @@ export class CalculateAge extends Expression {
 }
 
 export class CalculateAgeAt extends Expression {
-  precision:  ELM.DateTimePrecision;
-  
+  precision: ELM.DateTimePrecision;
+
   constructor(json: ELM.CalculateAgeAt) {
     super(json);
     if (!json.precision) {
-      throw new Error('Precision is required')
+      throw new Error('Precision is required');
     }
     this.precision = json.precision;
   }
--- a/src/elm/clinical.ts
+++ b/src/elm/clinical.ts
@@ -2,15 +2,17 @@ import { Expression } from './expression';
 import * as dt from '../datatypes/datatypes';
 import { Context } from '../runtime/context';
 import { build } from './builder';
+import ELM from '../types/elm'
 
 export class ValueSetDef extends Expression {
   name: string;
   id: string;
   version?: string;
 
-  constructor(json: any) {
+  constructor(json: ELM.ValueSetDef) {
+    // @ts-ignore
     super(json);
-    this.name = json.name;
+    this.name = json.name || '';
     this.id = json.id;
     this.version = json.version;
   }
@@ -28,10 +30,10 @@ export class ValueSetRef extends Expression {
   name: string;
   libraryName: string;
 
-  constructor(json: any) {
+  constructor(json: ELM.ValueSetRef) {
     super(json);
-    this.name = json.name;
-    this.libraryName = json.libraryName;
+    this.name = json.name || '';
+    this.libraryName = json.libraryName || '';
   }
 
   exec(ctx: Context) {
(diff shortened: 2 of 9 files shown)


src/elm/arithmetic.ts
```diff
@@ -381,7 +382,7 @@ export class Power extends Expression {
 }
 
 export class MinValue extends Expression {
-  static readonly MIN_VALUES = {
+  static MIN_VALUES = {
```
mgramigna, line 385: Is there a reason to take out the `readonly` qualifier here? The value of `MIN_VALUES` never gets re-assigned. I saw that `MAX_VALUES` below still has `readonly`, so figured I'd ask about the change

src/elm/arithmetic.ts
```diff
@@ -391,9 +392,13 @@ export class MinValue extends Expression {
 
   valueType: keyof typeof MinValue.MIN_VALUES;
 
-  constructor(json: any) {
+  constructor(json: ELM.MinValue) {
     super(json);
-    this.valueType = json.valueType;
+    if (json.valueType in MinValue.MIN_VALUES) {
+      this.valueType = json.valueType as keyof typeof MinValue.MIN_VALUES;
```
mgramigna, line 398: Small (opinionated) comment. Now that we use `keyof typeof MinValue.MIN_VALUES` in more than one place, maybe we can bring it out to its own `type` outside of the class? E.g.:

``` TypeScript
type MinValueType = keyof typeof MinValue.MIN_VALUES;
```

Then we can just re-use this type where applicable. Same applies to max values below

src/elm/arithmetic.ts
```diff
@@ -391,9 +392,13 @@ export class MinValue extends Expression {
...
+    } else {
+      throw new Error(`Not expecting MIN_VALUE: ${json.valueType})`);
```
mgramigna, line 400: This error check breaks the arithmetic tests, but for a very logical reason lol. There is a [test case](https://github.com/cqframework/cql-execution/blob/3b667a6f0cca2fd6d1b4772f9457f70325e866d0/test/elm/arithmetic/arithmetic-test.ts#L267) that asserts that any incorrect `MinValue` provided should throw an error _during execution_. The generated ELM for this test case is [here](https://github.com/cqframework/cql-execution/blob/3b667a6f0cca2fd6d1b4772f9457f70325e866d0/test/elm/arithmetic/data.js#L2799).

Since the constructor now throws an error if provided a `valueType` that we don't recognize.

I'm going to defer to @cmoesel here on how exactly to proceed. I think I might lean towards not throwing an error in a constructor, but I see the merit in the code that you've added.

src/elm/arithmetic.ts
```diff
@@ -422,9 +427,13 @@ export class MaxValue extends Expression {
...
   valueType: keyof typeof MaxValue.MAX_VALUES;
 
-  constructor(json: any) {
+  constructor(json: ELM.MaxValue) {
     super(json);
-    this.valueType = json.valueType;
+    if (json.valueType in MaxValue.MAX_VALUES) {
+      this.valueType = json.valueType as keyof typeof MaxValue.MAX_VALUES;
+    } else {
+      throw new Error(`Not expecting Max_VALUE: ${json.valueType})`);
```
mgramigna, line 435: Same deal here as my comment about errors for MinValue.

src/elm/clinical.ts
```diff
@@ -2,15 +2,17 @@ import { Expression } from './expression';
...
+import ELM from '../types/elm';
 
 export class ValueSetDef extends Expression {
   name: string;
   id: string;
   version?: string;
 
-  constructor(json: any) {
+  constructor(json: ELM.ValueSetDef) {
+    // @ts-ignore
```
mgramigna, line 13: Looks like this `ts-ignore` is because of issues with the ELM types? Is this a known issue? If so, we should maybe add a comment explaining the presence of this, and what the discrepancy of the ELM types are. To make sure `eslint` doesn't fail, you should also disable the lint rule above this line:

```suggestion
    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
    // @ts-ignore
```

src/elm/clinical.ts
```diff
@@ -2,15 +2,17 @@ import { Expression } from './expression';
...
     super(json);
-    this.name = json.name;
+    this.name = json.name || '';
```
mgramigna, line 15: After looking at the [logcial specification](https://cql.hl7.org/04-logicalspecification.html#valuesetdef) I actually think I defined the type of `name` incorrectly in this class. `name` is `0..1`, so I believe in the declaration of the class member, it should be `name?: string`. That should allow you to get rid of the `|| ''` assignment here. My bad!

src/elm/clinical.ts
```diff
@@ -28,10 +30,10 @@ export class ValueSetRef extends Expression {
   name: string;
   libraryName: string;
 
-  constructor(json: any) {
+  constructor(json: ELM.ValueSetRef) {
     super(json);
-    this.name = json.name;
-    this.libraryName = json.libraryName;
+    this.name = json.name || '';
+    this.libraryName = json.libraryName || '';
```
mgramigna, line 36: Same deal as above w/r/t the [logical specification](https://cql.hl7.org/04-logicalspecification.html#valuesetref). `name` and `libraryName` should be `name?: string` and `libraryName?: string` respectively, and these default empty strings can be removed. My bad again!

src/elm/clinical.ts
```diff
@@ -103,11 +111,12 @@ export class CodeSystemDef extends Expression {
   id: string;
   version: string;
 
-  constructor(json: any) {
+  constructor(json: ELM.CodeSystemDef) {
+    // @ts-ignore
```
mgramigna, line 115: Same as initial comment with more explanation of ts-ignore and disabling the lint rule

src/elm/clinical.ts
```diff
@@ -121,11 +130,12 @@ export class CodeDef extends Expression {
   systemName: string;
   display?: string;
 
-  constructor(json: any) {
+  constructor(json: ELM.CodeDef) {
+    // @ts-ignore
```
mgramigna, line 134: Same as initial comment with more explanation of ts-ignore and disabling the lint rule

src/elm/clinical.ts
```diff
@@ -183,14 +195,15 @@ export class ConceptDef extends Expression {
   codes: any;
   display?: string;
 
-  constructor(json: any) {
+  constructor(json: ELM.ConceptDef) {
+    // @ts-ignore
```
mgramigna, line 199: Same as initial comment with more explanation of ts-ignore and disabling the lint rule

src/elm/clinical.ts
```diff
@@ -158,11 +168,13 @@ export class Code extends Expression {
...
 
-  constructor(json: any) {
+  constructor(json: ELM.Code) {
     super(json);
     this.code = json.code;
-    this.systemName = json.system.name;
-    this.version = json.version;
+    this.systemName = json.system?.name || '';
+    // @ts-ignore
+    // TODO: Spec says version does not exist on code?
```
mgramigna, line 176: you're right.... @cmoesel ??

src/elm/clinical.ts
```diff
@@ -139,10 +149,10 @@ export class CodeRef extends Expression {
   name: string;
   library: string;
 
-  constructor(json: any) {
+  constructor(json: ELM.CodeRef) {
     super(json);
-    this.name = json.name;
-    this.library = json.libraryName;
+    this.name = json.name || '';
+    this.library = json.libraryName || '';
```
mgramigna, line 155: Ack. Same deal as the logical specification stuff above (https://cql.hl7.org/04-logicalspecification.html#coderef)

Looks like I improperly marked these as required for the class

src/elm/clinical.ts
```diff
@@ -158,11 +168,13 @@ export class Code extends Expression {
...
   display?: string;
 
-  constructor(json: any) {
+  constructor(json: ELM.Code) {
     super(json);
     this.code = json.code;
-    this.systemName = json.system.name;
-    this.version = json.version;
+    this.systemName = json.system?.name || '';
+    // @ts-ignore
```
mgramigna, line 175: add eslint comment

src/elm/clinical.ts
```diff
@@ -202,12 +215,12 @@ export class ConceptDef extends Expression {
 export class ConceptRef extends Expression {
   name: string;
 
-  constructor(json: any) {
+  constructor(json: ELM.ConceptRef) {
     super(json);
-    this.name = json.name;
+    this.name = json.name || '';
```
mgramigna, line 220: ugh you know the drill at this point. My fault again, property can be optional. It's almost like I should've read the dang [manual](https://cql.hl7.org/04-logicalspecification.html#conceptref) when doing this conversion. Sorry

src/elm/clinical.ts
```diff
@@ -217,34 +230,37 @@ export class Concept extends Expression {
...
-  exec(ctx: Context) {
+  exec(ctx: Context): dt.Concept {
     const codes = this.codes.map((code: any) => this.toCode(ctx, code));
     return new dt.Concept(codes, this.display);
   }
 }
 
 export class CalculateAge extends Expression {
-  precision: string;
+  precision: ELM.DateTimePrecision;
```
mgramigna, line 257: Oooo love that we get the specific literals now rather than having to use general string. Yay!

src/elm/clinical.ts
```diff
@@ -217,34 +230,37 @@ export class Concept extends Expression {
...
 
-  constructor(json: any) {
+  constructor(json: ELM.CalculateAge) {
     super(json);
+    if (!json.precision) {
+      throw new Error('Precision is required');
+    }
```
mgramigna, line 263: Hmm.. I wonder how to handle this case. It's marked as optional in the [type defs](https://github.com/csenn/cql-execution/blob/e7c3267eabba2a688617de3d735ba444b8f8478e/src/types/elm.d.ts#L2563), but we throw an error when it isn't provided. Does omitting this error break some assumptions we've made in this class? Based on the [logical specification](https://cql.hl7.org/04-logicalspecification.html#calculateage), looks like precision should be `0..1`. Ugh. This might snowball into types being sad in other functions, and I've now realized is yet again my fault. Let's discuss with @cmoesel what to do with this one.

And fine I'll come clean now. Yes, I just very recently realized the utility of the logical specification...

src/elm/clinical.ts
```diff
@@ -262,9 +278,13 @@ export class CalculateAge extends Expression {
...
 export class CalculateAgeAt extends Expression {
-  precision: string;
-  constructor(json: any) {
+  precision: ELM.DateTimePrecision;
+
+  constructor(json: ELM.CalculateAgeAt) {
     super(json);
+    if (!json.precision) {
+      throw new Error('Precision is required');
+    }
```
mgramigna, line 287: Same as above comment on precision optional

src/elm/clinical.ts
mgramigna: Notably though, this breaks [this test](https://github.com/cqframework/cql-execution/blob/3b667a6f0cca2fd6d1b4772f9457f70325e866d0/test/elm/clinical/clinical-test.ts#L168), and I think snowball effects into [this one](https://github.com/cqframework/cql-execution/blob/3b667a6f0cca2fd6d1b4772f9457f70325e866d0/test/elm/clinical/clinical-test.ts#L186) too

mgramigna: Oops, and [this one](https://github.com/cqframework/cql-execution/blob/3b667a6f0cca2fd6d1b4772f9457f70325e866d0/test/elm/clinical/clinical-test.ts#L209) and maybe [this one](https://github.com/cqframework/cql-execution/blob/3b667a6f0cca2fd6d1b4772f9457f70325e866d0/test/elm/parameters/parameters-test.ts#L188)

They all have to deal with a default assignment to version 🤔

mgramigna (commented):
Thanks for the pull request, @csenn! Happy to see ELM TypeScript types being explored and implemented. It's something that's been on my radar for a long time, but I've never put the effort in, so I'm very glad that you have :).

Also, thanks for adding some better non-ELM-related types to the repo! This was our vision, to start with a bunch of `any`s or non-specific types for the initial conversion, and slowly improve it as we go.

I've left a bunch of inline comments. Some are just for discussion, and some have direct action. Most probably require @cmoesel's thoughts too.

src/elm/clinical.ts
cmoesel: OK.  So here's the thing:

* The logical spec (and XSD) do allow precision to be missing.
* Neither say how you should calculate the age when it's missing.  Which is quite unfortunate.
* BUT... I don't think there is any way in the CQL _language_ to create this type _without_ a precision, because the language exposes it via specific [unit-based functions](https://cql.hl7.org/09-b-cqlreference.html#calculateage) (`CalculateAgeInYears`, `CalculateAgeInMonths`, etc); there is no `CalculateAge` without a built-in precision unit.
* So in some ways, this conversation is moot, because, in theory, precision will never _actually_ be undefined.

What does that mean for us here? I guess I don't know -- but probably it means we should just do whatever is easiest. ;-)

src/elm/arithmetic.ts
csenn: Nope, just added readonly back!

src/elm/arithmetic.ts
csenn: It also does make sense to pull MIN_VALUE out, and maybe even use an ENUM instead of a dict. But it was a little weird to even have to use `as keyof typeof MinValue.MIN_VALUES`... I thought Typescript would be able to infer that since `json.valueType in MinValue.MIN_VALUES` must be true

src/elm/arithmetic.ts
csenn: Yeah def agree with you, both ways have merit. Here is a [StackOverflow](https://stackoverflow.com/a/77705) discussing. I semi-lean towards validation in the constructor as long as that runs over the whole ELM tree as soon as the ELM wrapper is instantiated, just so that it doesn't become possible for execution to 'skip' certain parts of the ELM tree depending on the input patient data (only exec MinValue node if Patient has Condition Y) which could then allow bugs to escape tests more easily. Not sure if that's applicable though?

Either way, MIN_VALUES could be an ENUM but SHOULD NOT be added to the ELM generated from the XML because the mappings to `MathUtil.MIN_INT_VALUE` seem to be numbers that are only specific to the JS implementation.

src/elm/clinical.ts
csenn: So this is a tricky one. @cmoesel also including you here. The issue is that the `Def` classes ValueSetDef, ConceptDef, CodeSystemDef, etc. DO NOT inherit from `Expression` in the XML schema data. Yet the class in the TypeScript does.
```
<xs:complexType name="ValueSetDef">
<xs:complexContent>
	<xs:extension base="Element">
		<xs:sequence>
			<xs:element name="codeSystem" type="CodeSystemRef" minOccurs="0" maxOccurs="unbounded">
			</xs:element>
		</xs:sequence>
		<xs:attribute name="name" type="xs:string"/>
		<xs:attribute name="id" type="xs:string" use="required">
		</xs:attribute>
		<xs:attribute name="version" type="xs:string" use="optional">
		</xs:attribute>
		<xs:attribute name="accessLevel" type="AccessModifier" use="optional" default="Public"/>
	</xs:extension>
</xs:complexContent>
</xs:complexType>

export class ValueSetDef extends Expression {
  name: string;
  id: string;
  version?: string;

  constructor(json: ELM.ValueSetDef) {
    // @ts-ignore
    super(json);
    this.name = json.name || '';
    this.id = json.id;
    this.version = json.version;
  }
etc...
}

```
Notice the `base='Element'`. In most other "expressions", the XML looks has `base='Expression'`: 
```

<xs:complexType name="MinValue">
<xs:complexContent>
	<xs:extension base="Expression">
		<xs:attribute name="valueType" type="xs:QName" use="required"/>
	</xs:extension>
</xs:complexContent>
</xs:complexType>


```
To further complicate, the generated ELM for the `Def` expressions do not carry a `type` property that allows the differentiation of an expression type. 
```
// No Type
interface ValueSetDef extends AbstractElement {
      name?: string;
      id: string;
      version?: string;
      accessLevel?: AccessModifier;
      codeSystem?: CodeSystemRef[];
}

// Has Type
interface MinValue extends AbstractExpression {
      type: "MinValue";
      valueType: string;
}
```
Which is necessary to differentiate the different types of Expressions currently being evaluated using 
```   
type Expression =
      | OperatorExpression
      | ExpressionRef
      | FunctionRef
      | ParameterRef
      | + a bunch more
```
So to summarize, the problem could possibly be solved by simply adding 'base=Expression' to the def in the XML, but 

1. Not sure the Core Java Code wants us to keep changing the XML spec for this library (maybe through an MR)? There would then likely be changes to the generated ELM (adding of the type key for expression differentiation) which could potentially break things. Also that may not make logical sense anyways, because a `ValueSetDef` can not exist anywhere in the Expression tree, only at the root ELM.Library layer.
2. Could change this code a bit on the JS side, ensuring that these `Def` classes do not inherit from Expression. But of course we would need to think through that a bit more than just adding `// @ts-ignore` as I did here.

src/elm/clinical.ts
csenn: I personally don't hate defaulting to empty strings because usually you just do a check like ```
if (json.name) {
   // Do Something
}
```
and null or empty string behave the same way. It also makes the code and types simpler sometimes without having to always check for null in cases where you want to do something like 
```
json.name.toLowerCase()
```
But the broader discussion goes back to whether there are cases where we'd actually want to update the XML spec itself and change certain properties to required or not. There are pros and cons to having the single source of truth for the ELM spec. But it could get messy if we were to say manually start editing the generated ELM Typescript Types, because then we would not be able to generate it again later when the XML schema changes.

File: src/elm/clinical.ts (hunk 15 above)
Comment (csenn) [reply src/elm/clinical.ts:257]: Yeah this is nice. In theory we could also make this an ENUM instead of
```
  type DateTimePrecision =
    | 'Year'
    | 'Month'
    | 'Week'
    | 'Day'
    | 'Hour'
    | 'Minute'
    | 'Second'
    | 'Millisecond';
```
but both probably are fine.

src/elm/clinical.ts
csenn: I think this goes back to the previous comment about possibly updating the XML spec to ensure that precision is always there if it should always be there... Interested to hear both of your thoughts on this one. At worse we could always just create some special ELM types locally that act as overrides, something like `interface CqlExecCalculateAge extends ELM.Expression` but that will get annoying in more than just a few case

src/elm/arithmetic.ts
cmoesel: I can see both arguments -- including the desire to fail immediately upon loading the ELM (rather than on execution, just in case execution never occurs).  That said the spec does say this:
> For any other type, attempting to invoke minimum results in an error.

One could argue that "attempting to invoke" implies an error on execution, not on loading. (As far as I can tell, this is how the [Java cql-engine implementation](https://github.com/DBCG/cql_engine/blob/master/engine/src/main/java/org/opencds/cqf/cql/engine/elm/execution/MinValueEvaluator.java) behaves as well).

src/elm/clinical.ts
cmoesel: Or maybe we should just tell eslint to allow `@ts-ignore`?  If we know we need to use it in a bunch of places...

src/elm/clinical.ts
cmoesel: I think it makes sense to file an issue against CQL asking that this property either be made required, or that the behavior is defined for when a precision is not provided.

cmoesel:
Sorry I let this linger for so long... I've responded in some of the conversations with my thoughts.  Interested to hear yours,

src/elm/clinical.ts
csenn: Hey @cmoesel, interfaces would definitely be an interesting approach to explore. It makes sense that switching to TypeScript might lean the library towards certain design patterns that might have been handled a little differently in vanilla JavaScript. 

Sorry I haven't made any more progress on this MR, I've actually been working on another related project that you may find interesting though! It's a browser based CQL IDE and publisher tool that provides support for real-time validation of CQL, CQL statement type checking, dependency validation and resolution, and other tooling that supports writing CQL. It uses cql-exec under the hood for CQL execution. 

If you're interested, I'd love to share the progress and maybe get some feedback from you and @mgramigna in a week or two and see what you guys think about the approach?

P.S. - It also may be worth looking into the type support for the cql execution results. One thing I'm seeing with fhir-exec is the results are not in FHIR format. For instance, a codeable_concept.code has a `.value` property, which in FHIR is just a string.

cmoesel: Hi @csenn -- yes, your side-project does sound interesting! Will it be available via an open source license?  If so, cool, I'd love to see it.  If it won't be available via open source, I need to determine if it's appropriate for us to engage at that level (just because of our position in this space and engagement w/ other companies).

As for the results, it might be hard to type those since CQL allows you to return arbitrary Tuples.  I suppose we could type the FHIR result types and authors can cast to them if they're sure of what it is.  The reason for the variation from FHIR is that the result types align with the CQL FHIR model info (which expresses FHIR primitives as complex types with a `value`) rather than the FHIR JSON serialization.  I've thought in the past that it might be nice to add a `toFHIRJSON` function on the types to allow for conversion.

csenn: Ah interesting, that makes sense that it's the model info. A `toFHIRJSON` could likely be useful though... Maybe we can discuss how that could be best achieved and I can help look into it. 

I'm actually not 100% sure where the project I'm working on is going, more just trying to figure what makes sense right now. It will certainly be free, but probably a combination of open and closed source since there will need to be login, data persistence, CI pipeline, etc. If you do have a chance to take a quick gander once it's ready, you'll probably pretty immediately see why the output type matters. Since data IN is FHIR, it's a little hard to intuit that OUT comes a different data structure especially for common scenarios such as filtering Conditions by a ValueSet. But yeah like you mentioned, when shaping and returning Tuples you're actually not working with FHIR anymore, so it's tricky...

src/elm/arithmetic.ts
cmoesel: @csenn - I'm taking a look at this PR again, as it would be great to get it resolved. I noticed that you said you added the `readonly` back on line 385, but when I look at the current code on this branch, I don't see it. Perhaps you have not pushed it back up?

src/elm/clinical.ts
cmoesel: On re-review (in an effort to button this PR up), I think our guidance is to update this code to doc why we have to `@ts-ignore` it and to add the eslint rule as Matt suggested.  This is a potential point for improvement in the future, but let's just leave it at that for now.

src/elm/clinical.ts
cmoesel: At this point, unless there is a strong reason for it, I'd shy away from introducing changes to the ELM specification. Those changes are expensive given the governance process on CQL.  In cases where something isn't clearly wrong, it may be better just to accommodate for it.

cmoesel:
@csenn - I've given this another round of review.  I think there are a few small tweaks that would be good to make, but overall everything looks to be in good shape.  The question on whether or not to throw in the constructor is still unresolved.  I think it's probably OK to keep it as you have it in this PR, but I've also asked a colleague for additional input before making the call.

So... I think this branch needs to be brought up to date w/ master and a few very minor things (mainly doc/ts-ignore) need to be tweaked.  I realize that this PR has been open for some time now, so if you (@csenn) don't have time to revisit it, I could potentially make the changes myself if that is OK w/ you.  Let me know how you want to proceed.

mgramigna:
Agreed with @cmoesel on the constructor throwing. I think it's okay as is.

One thing I wanted to ask about was the intention of these ELM types going forward. I think we're happy to include the `elm.d.ts` in the cql-execution project, but were there hopes of getting this into [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) so we can use a `@types/elm` in any TypeScript project? Not entirely sure what the process of that would be, but safe to say I think we'd be happy to include the type declarations in the source code of `cql-execution` in the meantime

csenn:
Hey @cmoesel and @mgramigna, sorry this fell on the wayside. I used the ELM types generated in my own projects, and it's been very helpful for my needs, but the question is are the types good enough for DefinitleyTyped. Here is the repo I used to generate the types: https://github.com/cqlabio/cql-elm-types. 

It converts the XSD into Typescript types. Unfortunately, initially i tried to keep it simple and just use libraries that converted XSD into json_schema and json_schema into typescript types, however the types did not look good. So I ended up just writing some code that converts it manually, but it's not super well tested.

So either A) we can try and validate and possibly improve what I built or B) we can just create a new de-facto repo with the index.d.ts file and basically hand code changes for now on. What do you guys think? Does ELM change enough that automated would be better? Is that XSD reliable enough or does hand crafting have some advantages (such as marking something required that is missing in the XSD)? 

In terms of this PR, it would probably be better to just restart it since it only took 0.5 hours - 1 hour if I remember right. Is there a plan for any large architecture changes to what's currently on master/main?

cmoesel:
@csenn -- I think it makes sense to target DefinitelyTyped.  I wouldn't worry _too much_ about some types not being 100% perfect due to idiosyncrasies, etc; it's easy enough to fix them if we find anything like that.

As for automating type generation vs hand-maintaining from here forward... I feel like ELM is pretty stable at this point. If getting the automation solid requires a lot of hacks/workarounds or is just generally difficult, I feel like hand-maintaining the types should be quite feasible.  It also allows for some one-offs and judgement calls that might not be possible in automation.

All that said, before we decide how to move forward, I think it would be good to bring @brynrhodes into this conversation. Have you talked to him about this at all? @bkaney is also a big typescript fan who does some work in CQL, so he might have some feedback and tips as well.

cmoesel:
Closing this PR as it has become pretty stale at this point and @csenn indicated in his last comment that he might start a new PR anyway if he decides to continue pursuing this. That said, I'm happy to re-open the PR in the future if requested.

PR #270, opened 2022-06-03
Upgraded TSC config to target ES2017

hossenlopp:
Updates TypeScript compilation to target ES2017 standards. This allows for native use of newer ECMA Script standards.

Pull requests into cql-execution require the following.
Submitter and reviewer should ✔ when done.
For items that are not-applicable, mark "N/A" and ✔.

[CDS Connect](https://cds.ahrq.gov/cdsconnect) and [Bonnie](https://github.com/projecttacoma/bonnie) are the main users of this repository.
It is strongly recommended to include a person from each of those projects as a reviewer.

**Submitter:**
- [x] This pull request describes why these changes were made
- [x] Code diff has been done and been reviewed (it does not contain: additional white space, not applicable code changes, debug statements, etc.)
- [x] Tests are included and test edge cases N/A
- [x] Tests have been run locally and pass
- [x] Code coverage has not gone down and all code touched or added is covered.
- [x] Code passes lint and prettier (hint: use `npm run test:plus` to run tests, lint, and prettier)
- [x] All dependent libraries are appropriately updated or have a corresponding PR related to this change
- [x] `cql4browsers.js` built with `npm run build:browserify` if source changed.

**Reviewer:**

Name:
- [ ] Code is maintainable and reusable, reuses existing code and infrastructure where appropriate, and accomplishes the task’s purpose
- [ ] The tests appropriately test the new code, including edge cases
- [ ] You have tried to break the code

PR #271, opened 2022-06-23
Convert cql-execution to async

mgramigna:
~The behemoth has arrived. I'm putting this into draft, because I think #270 should get merged first, which greatly improves the performance of our async code.~

# Justification

**NOTE**: A lot of this justification might reference FHIR-specifc things. But the benefits of an asynchronous `cql-execution` engine theoretically apply to any data model that interfaces with `cql-execution`

The main motivation for converting  `cql-execution` to async was to allow data providers to implement methods that use asynchronous operations to retrieve patient data (e.g. HTTP calls, database lookups, etc.)
Currently, all of the `exec` functions for ELM expressions are synchronous, which in turn means that all of the data for a given patient needs to live in memory of the engine/data provider during execution-time of the ELM.

We've been doing some experimentation with [a data source](https://github.com/projecttacoma/cql-exec-fhir-mongo/) that uses MongoDB to retrieve patient data at runtime from a given Mongo database, as we think this could lead to some cool implementations of data providers for `cql-execution`, and some cool performance optimizations for a large amount of patient data.

The way [cql-exec-fhir](https://github.com/cqframework/cql-exec-fhir/) is written right now, it requires all of the Patient Bundles to be passed in to the `PatientSource` class
before execution happens. This can quickly cause the JavaScript heap to run out of memory when we reach a vast amount of patients.

The burden of migrating to an asynchronous usage from a user-perspective is low: just use `async/await` or `.then`

# Summary

This PR does two main things:

1. Convert all ELM expressions' `exec` methods to be async. This is necessary because we cannot make assumptions about where in the ELM expression tree asynchronous calls might happen. Basically, since a `Retrieve` (which could now be doing async stuff)
could be anywhere in the tree, the whole tree should be able to bubble up the `Promise` it returns and await it
2. Convert calls to `findRecords`/`currentPatient`/`nextPatient` to async. This keeps backwards compatibility with existing data sources, while also allowing new data sources to return `Promise`s instead of just raw data

Because of a combo of 1. and 2., the main `Executor` class that is exposed by `cql-execution` needs to be called in an asynchronous pattern as well

# Code Changes

* Update all `src/elm/*` to use `async` functions
* Add `await` to any code that uses the above functions (including in tests)
* Added smart detection of `Promise`s from `currentPatient`/`nextPatient` iteration and await when a `Promise` is returned by the data source for those functions
* Update `src/runtime/context.ts` to use async `findRecords` calls
* Update examples and documentation to reflect async pattern and added migration guide document

# Testing Guidance

* Run all tests and run examples in the `examples` directory to see that the results are the same as the synchronous version
* For proof that asynchronous patient sources work, see this [example repo](https://github.com/mgramigna/cql-exec-async-example) I created that uses `cql-execution` and a fork of `cql-exec-fhir` to pull patient data from a FHIR server

# Advice

Have fun

---

**Submitter:**

**Reviewer:**

Name:


test/elm/external/external-test.ts
```diff
@@ -14,83 +14,83 @@ describe('Retrieve', () => {
     findRecordsSpy = sinon.spy(this.ctx, 'findRecords');
   });
 
-  it('should find conditions', function () {
-    const c = this.conditions.exec(this.ctx);
-    c.should.have.length(2);
+  it('should find conditions', async function () {
+    const c = await this.conditions.exec(this.ctx);
+    should(c).have.length(2);
```
cmoesel, line 19: Just curious -- why the change from `c.should` to `should(c)`? I've just tested both ways and they seem to work the same.  Is this just a preference on your part or am I missing something?  Is that in case `c` comes back `undefined` for some reason?

test/util/util-test.ts
```diff
@@ -24,3 +25,87 @@ describe('typeIsArray', () => {
...
+  });
+
+  it('should return false for non-promise', () => {
+    isPromise('not a promise').should.be.false();
+  });
+
+  it('should return true for promise', () => {
+    isPromise(Promise.resolve('definitely a promise')).should.be.true();
+  });
+});
```
cmoesel, line 111: If you kill `isPromise` you can obviously kill these tests too.

V2_to_V3_MIGRATION.md
```diff
@@ -0,0 +1,25 @@
+This document outlines the breaking changes contributed to `cql-execution` that will be included in a `v3.0.0` release of the library. This document will be updated as new breaking changes are contributed to the repository before finalizing a `v3.0.0` release.
+
+## Async Conversion
+
+[This pull request](https://github.com/cqframework/cql-execution/pull/271) converted the core execution code of `cql-execution` to be asynchronous. This enables a wider variety of [DataProviders](https://github.com/cqframework/cql-execution/blob/7ecb00b236903fc0816966e4ca8368d50d6afbc4/src/types/cql-patient.interfaces.ts#L8)
+to integrate with `cql-execution`, as data can now be retrieved using asynchronous operations (e.g. HTTP requests, database lookups, etc.). This conversion requires a change in how an [Executor](https://github.com/cqframework/cql-execution/blob/master/src/runtime/executor.ts) is used in practice:
```
cmoesel, line 6: Now that we support async `TerminologyProvider` implementations too, it may be worth noting that here.

cmoesel (commented):
This is really great work, @mgramigna!  It's tedious, but important!  I'd still like to do some more testing with it, but I've done a pass through the code and left a few comments.  They're all very minor and should be pretty simple to resolve.

Once I've had a chance to more thoroughly test this w/ some real-world implementations, I'll give my official approval.

Regarding branches, I'd suggest that we update `next` to point to the HEAD of `master` (since `next` doesn't have any of its own commits yet), then rebase this on `next`. I think you have the proper permissions to do that, but let me know if you find otherwise.

test/elm/external/external-test.ts
mgramigna: 100% honesty: I don't fully know. This was a long time ago haha. I traced it back to https://github.com/projecttacoma/cql-execution/pull/18/commits/6e4fe50601059d88b71b804453fca2699dcc54ac

IIRC, I remember during the initial conversion, seeing a lot of tests failing due to "cannot read property "should" of undefined" which wasn't that helpful in tracing down the error. Wrapping it in the `should` function  call like this allowed for more useful messages like "expected <Promise> to have property `length`"? Short answer is that the tests gave better error messages when `c` would be null or undefined or something. Which I think helped trace down some errors in the conversion, but I'm not convinced it was actually necessary in all of these cases, oops

test/util/util-test.ts
mgramigna: RIP!

src/util/util.ts
```diff
@@ -74,3 +74,39 @@ export function getTimezoneSeparatorFromString(string: string) {
...
+      }
+    } else {
+      const sortedElem = right.shift();
+      if (sortedElem !== undefined) {
+        sorted.push(sortedElem);
+      }
+    }
+  }
+  return [...sorted, ...left, ...right];
+}
```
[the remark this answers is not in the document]
mgramigna: Writing this reminded me about the hour long video I had to watch in driver's ed about how to deal with Moose. It really made me think that I would be encountering Moose far more often in my daily life.

I really did think I would be encountering merge sort so much more in my daily life when I was a wee CS freshman. I'm still baffled that doing this was the most logical solution based on this incredibly specific use case

test/elm/external/external-test.ts
cmoesel: OK.  No need to change it.  I was just curious.

src/elm/type.ts
```diff
@@ -485,9 +485,9 @@ export class ConvertsToTime extends Expression {
   }
 }
 
-function canConvertToType(toFunction: any, operand: any, ctx: Context) {
+async function canConvertToType(toFunction: any, operand: any, ctx: Context) {
   try {
-    const value = new toFunction({ type: toFunction.name, operand: operand }).execute(ctx);
+    const value = await new toFunction({ type: toFunction.name, operand: operand }).execute(ctx);
```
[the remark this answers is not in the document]
mgramigna: Agreed! Address in https://github.com/cqframework/cql-execution/pull/271/commits/ee246da1c3e7902889426da44d9c1629f9cebb1d with some bonus type safety on the conversion class arg ;)

test/elm/external/external-test.ts
mgramigna: Something I just realized in addition to this is that using the `should` wrapper gives more type safety? For instance, in the old statement `c.should.have.length(2)`, `c` has type any, and so this relies on the developer to understand how you chain together `should` assertions. When using the `function`, the TypeScript LSP can do smart autocomplete:

image

image

Nice little boost to developer experience IMO

There might be a way to get the same thing with better configuration on using `should` with TypeScript, but not sure how that would work

test/elm/aggregate/aggregate-test.ts
```diff
@@ -12,17 +12,17 @@ describe('Count', () => {
...
-  it('should be able to count empty list', function () {
-    this.empty.exec(this.ctx).should.equal(0);
+  it('should be able to count empty list', async function () {
+    (await this.empty.exec(this.ctx)).should.equal(0);
   });
-  it('should be able to count null list', function () {
-    this.is_null.exec(this.ctx).should.equal(0);
+  it('should be able to count null list', async function () {
+    (await this.is_null.exec(this.ctx)).should.equal(0);
   });
```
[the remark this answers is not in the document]
mgramigna: Yeah I'm not opposed to this! There's also:

``` TypeScript
return this.not_null.exec(this.ctx).should.be.resolvedWith(5);
```

Which compares the value using deep equality, so that fixes the stack trace issue and avoids the need for async await pattern in tests? That would also require changing the assertion everywhere just for the sake of using Promises but idk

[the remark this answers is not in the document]
mgramigna: https://shouldjs.github.io/#assertion-fulfilledwith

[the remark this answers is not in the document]
cmoesel: I totally forgot about that style assertion.  Good call.  It's not as succinct as my handy little functions, but I like that it will trace to the right place and it's probably a more correct way of using the framework.  I don't think we should do anything right now (unless you have some crazy emacs thing that can do it in > 1 hr) -- but let's keep this in mind.

test/elm/external/external-test.ts
cmoesel: Oh, that is nice. I don't know if I noticed it _not_ working before -- but seeing it working makes me want it.

test/elm/aggregate/aggregate-test.ts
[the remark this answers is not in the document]
mgramigna: I'm not saying I don't have the capability to use a macro that could accomplish this..... but I agree it probably makes sense to hold off and keep in mind for future refactors of the code

cmoesel (approved):
OK! This all looks splendid. Code is clean, regressions run fine.  We're all good!

Thanks so much for this @mgramigna and team!  As a close friend of mine once said:
> That's one small step for man, one giant leap for mankind.

PR #275, opened 2022-08-31
Fix AgeInYears/AgeInMonths when birthday is today

### LINKED ISSUE #273: AgeInYears() is uncertainty when birthday is today
If you execute `AgeInYears()` on a FHIR patient whose birthday is today, the result will be an uncertainty ranging from their previous age to their current (new) age.  The result should be their new age (not an uncertainty).

This bug was introduced in CQL Execution 2.4.1.  Prior to 2.4.1, the `ToDateTime(Date)` operation set all time components to `0`. When `AgeInYears()` was executed, it converted the birth Date to a DateTime (at midnight) and calculated _duration_ from `Now()`. This resulted in the expected values.

In 2.4.1, the `ToDateTime(Date)` operation was updated to set all time components to `null` (per the spec). As a result, when `AgeInYears()` is executed, the converted birth DateTime has null time components, so when doing a _duration_ against `Now()`, the result is an uncertainty (because it depends on what time the patient was born).

Looking more closely at the spec, the [Age operators](https://cql.hl7.org/09-b-cqlreference.html#age) section says this:
> Note that for AgeInYears and AgeInMonths, the birthDate is specified as a Date and Today() is used to obtain the current date; whereas with the other precisions, birthDate is specified as a DateTime, and Now() is used to obtain the current DateTime.

The CQL Execution engine does not follow this guidance (hence the bug above).  If we implement this, it should fix the problem. That said, the `AgeInX` operators are not surfaced directly in the ELM. They are surfaced as `CalculateAge` operators.  Although the [CalculateAge reference](https://cql.hl7.org/09-b-cqlreference.html#calculateage) doesn't include the same language as the `AgeInX` reference, we must assume that the guidance applies here since this is the only way to do `AgeInX` in ELM.

cmoesel:
This PR updates the `CalculateAge` logic to perform calculations using `Date`-based birthday and `Today()` when the precision is years or months.  This is based on the following language in the [spec](https://cql.hl7.org/09-b-cqlreference.html#age):

> Note that for AgeInYears and AgeInMonths, the birthDate is specified as a Date and Today() is used to obtain the current date; whereas with the other precisions, birthDate is specified as a DateTime, and Now() is used to obtain the current DateTime.

Prior to this change, results for `AgeInYears` and `AgeInMonths` could be uncertainties when the birthday is today (since it was considering time components in the calculation).

As part of this change, I've updated some of the existing tests to use a fixed execution datetime, as that makes the tests more reliable (and insusceptible to DST issues). This allowed me to also unskip some previously skipped tests. I've also introduced new tests to ensure the behavior is as expected when using year and month precision.

Fixes #273 

**Submitter:**

**Reviewer:**

Name:

### DIFF
--- a/examples/browser/cql4browsers.js
+++ b/examples/browser/cql4browsers.js
@@ -3877,15 +3877,20 @@ class CalculateAge extends expression_1.Expression {
         this.precision = json.precision;
     }
     exec(ctx) {
-        const date1 = this.execArgs(ctx);
-        const date2 = dt.DateTime.fromJSDate(ctx.getExecutionDateTime());
-        const result = date1 != null ? date1.durationBetween(date2, this.precision.toLowerCase()) : undefined;
-        if (result != null && result.isPoint()) {
-            return result.low;
+        const birthDate = this.execArgs(ctx);
+        // From the spec: "Note that for AgeInYears and AgeInMonths, the birthDate is specified as a
+        // Date and Today() is used to obtain the current date; whereas with the other precisions,
+        // birthDate is specified as a DateTime, and Now() is used to obtain the current DateTime."
+        // See: https://cql.hl7.org/09-b-cqlreference.html#age
+        let asOf;
+        if (this.precision.toLowerCase() === dt.DateTime.Unit.YEAR ||
+            this.precision.toLowerCase() === dt.DateTime.Unit.MONTH) {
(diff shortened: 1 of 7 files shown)


hossenlopp (approved):
Looks good and is thoroughly tested. New tests showed the errors with the old code. Tested in fqm-execution with not issues. Approved from the Abacus FHIR team aspect.

dczulada (approved):
Looks good

PR #298, opened 2022-11-03
Update DataProvider and PatientObject to allow async and be stricter

mgramigna:
* Make currentPatient/nextPatient return `PatientObject`s and allow Promises
* Make findRecords allow promises 

**Submitter:**

**Reviewer:**

Name:

### DIFF
--- a/src/types/cql-patient.interfaces.ts
+++ b/src/types/cql-patient.interfaces.ts
@@ -6,8 +6,8 @@ import { AnyTypeSpecifier } from './type-specifiers.interfaces';
  * Iterator for the patients provided to the execution engine
  */
 export interface DataProvider {
-  currentPatient(): any;
-  nextPatient(): any;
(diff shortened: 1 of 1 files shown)


mgramigna:
> @mgramigna Dude, why are you trying to make this slower by having patient resolved async (presumably via some http request). Its slow enough already.

The idea is to allow data providers to the engine to be asynchronous if they want to be. They don't have to be. Some ideas that were shared with us for this were not only HTTP requests but also database calls or something like that so you don't need to load all the patient data in memory in order to do execution, which might not scale well for some cases. We're looking for feedback and testing on the async stuff, that's why it's in beta.

No one is forced to implement async patient providers, and the worst thing that will happen is the engine might `await` something that isn't actually a Promise, which should be a negligible performance hit if any at all. I'm happy to test this statement more though and may even be proven wrong.

With the test cases that my team ran using an async CQL engine, we did not notice any slowdowns.

PR #306, opened 2023-01-15
Null sourced queries

cmoesel:
Fix handling of null-sourced queries so they return null as described here:
https://jira.hl7.org/browse/FHIR-40225

Also refactor multi-source queries to be more efficient, so we avoid executing sources multiple times.

And re-enable and improve some skipped tests.

Fixes https://github.com/projecttacoma/fqm-execution/issues/172

**Submitter:**

**Reviewer:**

Name:

### DIFF
--- a/src/elm/query.ts
+++ b/src/elm/query.ts
@@ -206,7 +206,7 @@ class AggregateClause extends Expression {
 }
 
 export class Query extends Expression {
-  sources: MultiSource;
+  sources: AliasedQuerySource[];
   letClauses: LetClause[];
   relationship: any[];
   where: any;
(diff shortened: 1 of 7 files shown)


src/elm/query.ts
```diff
@@ -237,8 +237,33 @@ export class Query extends Expression {
   }
 
   exec(ctx: Context) {
+    let sourceResults = this.sources.map(s => s.expression.execute(ctx));
+    // If every source is null, the result is null
+    // See: https://jira.hl7.org/browse/FHIR-40225
+    if (sourceResults.every(r => r == null)) {
+      return null;
+    }
+    const isList = sourceResults.some(r => Array.isArray(r));
```
mgramigna, line 246: Should we re-use the `typeIsArray` util instead?

src/elm/query.ts
```diff
@@ -237,8 +237,33 @@ export class Query extends Expression {
...
+    // For ease of processing, convert all the sources to lists and
+    // convert nulls to empty lists (we already handled the ALL null case)
+    sourceResults = sourceResults.map(r => {
+      if (r == null) {
+        return [];
+      } else if (Array.isArray(r)) {
```
mgramigna, line 252: Same as above. Not sure if this gives us any added benefit or not except consistency.

mgramigna (commented):
Can confirm that this fixes the issue in fqm-execution (and some bonus cases as well after clarifying the spec further!)

I just have one small question, but LGTM overall

src/elm/query.ts
cmoesel: These days, I think `Array.isArray` is pretty well-supported, so we should probably consider moving all uses of `typeIsArray` to `Array.isArray`.  Starting with this one. ;-)

src/elm/query.ts
cmoesel: Nah.

mgramigna (approved):
Thanks! Glad we were able to clarify this

birick1 (approved):
Ugh ... O[n^2] operations for cartesian products.  No better suggestion.

PR #308, opened 2023-03-06
Add Detailed Execution Error Reporting

natjoe4:

## Summary
Errors thrown during execution will now be throw as AnnotatedErrors, providing the Expression name, ELM local id, source Library identifier and version, as well as the initial error message. These changes aim to help faster identify where and why issues during execution arise.

### Code Changes
Added try/catch workflow to base Expression class on execute function that catches errors and throws AnnotatedErrors with additional execution failure information.

**Submitter:**

**Reviewer:**

Name:


mgramigna (commented):
Looks good! I was able to replicate errors locally both through fqm-execution and in cql-execution directly.

I'll lean on @cmoesel for more thoughts on the overall approach, I just had a few suggestions on a first glance

src/util/customErrors.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * A custom error that holds additional information and a more explicit message
+ * to simplify tracking down errors that occur during execution
+ */
+export class AnnotatedError extends Error {
+  constructor(message: string, expressionName: string, libraryName: string, localId?: string) {
+    super(
+      `Expression ${expressionName} from CQL Library: ${libraryName} ${
```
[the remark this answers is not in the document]
cmoesel: I wouldn't have given it much thought myself, but now that you got me thinking... Yeah, a little more structured output might be nice. I went even a step further:

```
AnnotatedError: Encountered unexpected error during execution.

	Error Message:	You done messed up.
	CQL Library:	AgeAtMP|1
	Expression:	CalculateAgeAt
	ELM Local Id:	12345

    at CalculateAgeAt.execute (/Users/cmoesel/git/cqframework/cql-execution/lib/elm/expression.js:40:19)
    at /Users/cmoesel/git/cqframework/cql-execution/lib/elm/expression.js:48:45
```

You don't have to use that, but if you want to, here's the code:
```ts
super(
  `Encountered unexpected error during execution.\n\n\tError Message:\t${message}\n\tCQL Library:\t${libraryName}\n\tExpression:\t${expressionName}${
    localId ? `\n\tELM Local Id:\t${localId}` : ``
  }\n`
);
```

cmoesel (changes requested):
This will be a really nice addition to the library. I only wish we had added it sooner!

I have a few suggestions for your consideration. Also, I noted that this PR is against `master`, but I believe that fqm uses the `next` branch, right? Should we direct it at `next` instead? I'm hoping that we'll agree to merge `next` to `master` soon and make 3.0 official.

cmoesel:
I noticed that the stack trace always points to the `execution` function in `Expression` -- which isn't terribly helpful. I went down a bit of a rabbit hole because TS would not recognize the `cause` property that I _thought_ would be inherited from `Error`. In the end, it turns out that if you just throw on a `cause` property yourself, it works as needed.

image

results in:

image

That's probably not particularly helpful or meaningful to an end user -- but it might be helpful to us.

src/elm/expression.ts
```diff
@@ -46,6 +55,20 @@ export class Expression {
...
     }
   }
+
+  /**
+   * Function used in error reporting during execution. Retrieves the source library from
+   * the context if it exists, or recursively traverses the context's parents until a source
+   * library identifier and version are found.
+   */
+  getRecursiveLibraryIdentifier(ctx: Context): string {
+    if (ctx instanceof PatientContext || ctx instanceof UnfilteredContext) {
```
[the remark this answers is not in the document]
mgramigna: Makes sense. Only annoying thing about this is since `ctx` is of type `Context` (which does not have a `.library` property), TypeScript gets mad because `library` doesn't exist on `ctx`. I could cast `ctx` to `any` both in the if and in the actually code for accessing the identifier.

I do agree that it would be nice to write the `if` in a more maintainable way that doesn't scale linearly with each new context added. Let me know if the cast is sufficient

mgramigna:
> I noticed that the stack trace always points to the `execution` function in `Expression` -- which isn't terribly helpful. I went down a bit of a rabbit hole because TS would not recognize the `cause` property that I _thought_ would be inherited from `Error`. In the end, it turns out that if you just throw on a `cause` property yourself, it works as needed.

Great find. Sounds like it's exactly what we'd want based on the scenario. From https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause:

> The cause data property of an [Error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error) instance indicates the specific original cause of the error.
>
> It is used when catching and re-throwing an error with a more-specific or useful error message in order to still have access to the original error.

Now that the original error is passed in to the `AnnotatedError` constructor, is it even worth having an individual `message` parameter in that constructor now? All we initialize for that is e.message. I could keep it for flexibility, or just remove it entirely?

src/util/customErrors.ts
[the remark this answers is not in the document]
mgramigna: Pushed this change, but I forgot to update the unit tests. Obviously I could just update the expected messages to have the new format. Is it worth making this more maintainable though? Maybe using a regex in the assertion instead to capture the specific pieces of information that we expect to be in the error, rather than a strict assertion on the entire string itself?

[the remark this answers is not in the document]
mgramigna: Potential alternative approach to testing:

```ts
try {
  testExpression.execute(this.ctx);
  fail();
} catch (e: any) {
  should(e).be.instanceOf(AnnotatedError);
  should(e.message).containEql('Error Message:\tExecution Failed!');
  should(e.message).containEql('CQL Library:\tTestSnippet|1');
  should(e.message).containEql('Expression:\tFailingExpression');
  should(e.message).containEql('ELM Local ID:\ttest');
}
```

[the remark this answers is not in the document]
cmoesel: > Is it worth adding a message if the localId is undefined suggesting enabling localIds in ELM translation to better locate the error? 

I don't have a strong opinion on this. It partially depends on how likely it is that the user who sees the error has control over the CQL-to-ELM translation process... Personally, I think I'd just leave it as it is -- and if people contact us for help and the expression use is ambiguous, we could ask them to recompile w/ local id.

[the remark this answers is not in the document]
cmoesel: > Maybe using a regex in the assertion instead to capture the specific pieces of information that we expect to be in the error, rather than a strict assertion on the entire string itself?

I think that's a fine approach -- and one that we use in my other projects. Sometimes, if I want to be sure the full message structure is correct, then I'll use the full message in _one_ unit test but the targeted regexes for all the others. That way if you intentionally tweak the message, there is only one unit test to fix; but if you _unintentionally_ tweak it, you still get a failed test to let you know.

src/elm/expression.ts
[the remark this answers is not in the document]
cmoesel: Yeah, I think casting is sufficient. If you want to be fancy, you could create a throw-away interface like:
```ts
interface ContextWithOptionalLibrary extends Context {
  library?: Library;
}
```
and cast to that. Then you still get all the type-checking since it knows what `library` is.

src/util/customErrors.ts
```diff
@@ -0,0 +1,25 @@
...
+  // TODO: need to explicitly specify cause in TypeScript versions < 4.7
+  // Remove this when TypeScript is upgraded, and add the cause directly to the `super` call
+  cause: Error;
+
+  constructor(
+    originalError: Error,
+    message: string,
+    expressionName: string,
+    libraryName: string,
+    localId?: string
```
hossenlopp, line 15: Suggest adding the `locator` as a possible argument to AnnotatedError. This would allow a consumer of the error to print out the CQL around the error and highlight that clause that had the issue.

hossenlopp (commented):
Suggested useful addition. Otherwise this looks really good.

src/util/customErrors.ts
mgramigna: This is super nice. It requires adding a `locator` property on the root `Expression` class constructor, but I don't see any downside in doing that as long as the error reporting behaves properly when locators aren't there (similar to when localIds aren't there.

@cmoesel are you good with this potential change? basically:

```ts
// In Expression constructor
if (json.locator != null) {
  this.locator = json.locator;
}
```

and then rendering it in annotated errors:

```
        Error Message:  you messed up 
        CQL Library:    lib|5.0.000
        Expression:     CalculateAgeAt
        ELM Local ID:   28
        CQL Locator:    35:29-35:86
```

mgramigna: @hossenlopp You mentioned having consumers be able to utilize the information. This made me wonder if the `AnnotatedError` should also contain this rendered info in a structured way to potentially consume elsewhere?

```ts
try {
  // something
} catch(e: unknown) {
  if (e instanceof AnnotatedError /* imported from cql-execution? */) {
    // e has the following properties
    e.expressionName;
    e.libraryName;
    e.localId;
    e.locator;
  }
}
```

Could be overkill. Just a thought

hossenlopp: Isn't this data included in the error that gets thrown? But yeah exporting the type would allow it to be used without crazy shenanigans.

mgramigna: Technically it isn't, it's just placed directly in the error message. If we make the constructor args actually persist on the class instance (like using the `public` keyword in TS), then they will be available:

image

hossenlopp: I completely missed that they were not being persisted. But I do see benefit in doing so.

cmoesel: I'm down w/ all of this (locators in the constructor, stored properties, etc.).

Teamwork is the dream work!

mgramigna:
@cmoesel @hossenlopp Thanks for the detailed feedback! I believe everything above should be resolved. I'm happy with the current state of it -- let me know what you think!

src/util/customErrors.ts
```diff
@@ -0,0 +1,26 @@
...
+ */
+export class AnnotatedError extends Error {
+  // TODO: need to explicitly specify cause in TypeScript versions < 4.7
+  // Remove this when TypeScript is upgraded, and add the cause directly to the `super` call
+  cause: Error;
+
+  constructor(
+    originalError: Error,
+    message: string,
+    public expressionName: string,
```
hossenlopp, line 13: Wait, this `public name` make them instance properties without having to re-iterate yourself with `this.name = name` on all the things?

test/elm/expression/expression-test.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    try {
+      testExpression.execute(this.ctx.childContext());
+      fail('test should have thrown an error before reaching this statement');
+    } catch (e: any) {
+      should(e).be.instanceOf(AnnotatedError);
+      should(e.message).containEql('Error Message:\tExecution Failed!');
+      should(e.message).containEql('CQL Library:\tTestSnippet|1');
+      should(e.message).containEql('Expression:\tFailingExpression');
+      should(e.message).containEql('ELM Local ID:\ttest-nested');
```
hossenlopp, line 47: Should also check the public instance properties for the localId.

hossenlopp (changes requested):
Looks good. Just have one TS question and suggested test improvments.

Also I see the question of where to target this doesn't appear to be addressed. Will we be merging into `master` and then rebasing to use on `next`?

src/util/customErrors.ts
hossenlopp: 😎

mgramigna:
> Will we be merging into master and then rebasing to use on next?

@hossenlopp I think so yeah, shouldn't be too bad, might just need to use a slightly different assertion to test for promise rejection in one of the tests. But I think 2.X versions of the library would greatly benefit from this change as well

src/util/customErrors.ts
```diff
@@ -0,0 +1,26 @@
...
+    public libraryName: string,
+    public localId?: string,
+    public locator?: string
```
cmoesel, line 16: Now that these are properties, they're getting printed out when the error bubbles up:
image

This is fairly redundant and made me wonder if we should stop putting them in the message. 

One one hand, if we keep them in the message, then if they're caught in another app using this lib, and that app logs only the message, you still get all the data.

On the other hand, if we let the message just be the message (without all the supplemental data), then we allow consuming apps to make their own decision about how to display the error and associated data to the user (since they can get at it as discrete values). In a web app, this would allow for nicer formatting than our pre-formatted text message. (Although I suppose they could still do that by _ignoring_ `e.message` and using `e.cause.message` instead).

What do you think?

cmoesel (commented):
This is good stuff. I've added a few small suggestions and a question. TBH, none of them are PR-stoppers though; so let me know if you're not convinced regarding the suggestions.

src/elm/expression.ts
```diff
@@ -46,6 +68,20 @@ export class Expression {
...
   }
+
+  /**
+   * Function used in error reporting during execution. Retrieves the source library from
+   * the context if it exists, or recursively traverses the context's parents until a source
+   * library identifier and version are found.
+   */
+  getRecursiveLibraryIdentifier(ctx: Context): string {
+    if ((ctx as Context & { library?: Library }).library?.source?.library?.identifier) {
+      const identifier = (ctx as Context & { library: Library }).library.source.library.identifier;
```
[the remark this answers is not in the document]
mgramigna: I like this better!

src/util/customErrors.ts
```diff
@@ -0,0 +1,26 @@
...
+ * to simplify tracking down errors that occur during execution
+ */
+export class AnnotatedError extends Error {
+  // TODO: need to explicitly specify cause in TypeScript versions < 4.7
+  // Remove this when TypeScript is upgraded, and add the cause directly to the `super` call
+  cause: Error;
+
+  constructor(
+    originalError: Error,
+    message: string,
```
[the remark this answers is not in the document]
mgramigna: Haha forgot about that question! Cool, this is what I would prefer as well

src/util/customErrors.ts
mgramigna: What if we removed it from the message, but maybe some type of primer about how additional info may be available for annotated errors.

Should we document this somehow as well? Maybe in a debugging section in the README or something? I can see how these properties could be hidden by someone who doesn't know they're supposed to be there, so I think some type of indication that _some_ additional info is there in the message would be nice.

cmoesel: I honestly don't know what the right answer is here. But yeah, if we choose to go w/ the simple message and leave display of the details to the consumer, then I think it makes a lot of sense to document it. Good idea.

Do you think we should do it?

cmoesel: OK. We just agreed over slack that since we're both conflicted on this, it's best to go with the path of least resistance -- which is to leave it as it is now.

cmoesel (approved):
image

PR #320, opened 2024-07-11
Implement Implies Operator and update to cql2elm 3.12.0

hossenlopp:
Implements the `implies` operator and updates CQL translator to 3.12.0. Also updates gradle wrapper to latest 8.x for Java 21 support. Enables the spec-tests for Implies operator.

Note: The translator has a bug with choice types in the output that caused issues with some tests. This issue was manually edited out in the ELM after running the `build:test-data` command. https://github.com/cqframework/clinical_quality_language/issues/1380

**Submitter:**

**Reviewer:**

Name:


test/generator/src/main/java/JavaScriptTestDataGenerator.java
```diff
@@ -32,7 +32,7 @@
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-import javax.xml.bind.*;
+//import javax.xml.bind.*;
```
cmoesel, line 35: It looks like this line can be deleted.

cmoesel (commented):
This is a _drive-by_ review. I did not look at it closely (sorry), but I wanted to make one suggestion. Given that you had to hand edit some of the generated JSON for tests to pass, it _might_ be helpful to put a comment in the source code near the tests that it affects. I can picture myself updating the translator in the future, re-running the `build:test-data` command, getting broken tests, and being confused as to why they're breaking (because, of course, I will forget that you needed to hand-edit the JSON). So a little comment might help forgetful people like me to remember: _If this test breaks, check the generated JSON for a repeated `type` element._

BTW -- nice job figuring out the mysteries of the translator 2.x to 3.x changes so you could get the test generator working again. That's not fun. Thanks for taking it on!

p9g:
Probably should update https://github.com/cqframework/cql-execution/blob/master/CQL_Execution_Features.xlsx

test/elm/list/list-test.ts
```diff
@@ -671,37 +671,37 @@ describe('First', () => {
 
   it('should get first of a list of numbers and the localId should exist on the root context', async function () {
     (await this.numbers.exec(this.ctx)).should.equal(1);
-    should(this.ctx.localId_context[6]).not.be.undefined();
+    should(this.ctx.localId_context[213]).not.be.undefined();
```
cmoesel, line 674: Oh man. Fixing these must have been pretty annoying.

Also... technically speaking, `localId` is a _string_ -- and when we use it to add a key to the `localId_context`, we keep it a string (see [setLocalIdWithResult](https://github.com/cqframework/cql-execution/blob/implies-operator/src/runtime/context.ts#L174-L180)). So we probably _should_ access it using a string key as well (e.g., `this.ctx.localId_context['213']`). I guess it works if you pass in a number anyway (because javascript), but it leads to confusion since it makes it look like an array accessor. I know this is totally _not your fault_ (it was already like this), but... what do you think of fixing it?

Or... do you think we should consider creating a helper function to grab the right local id from the data file? It looks like we could define a function like this:
```ts
function getListLocalId(testName: string, statementName: string) {
  return data[testName].library.statements.def.find((d: any) => d.name === statementName).expression
    .source.localId;
}
```
and then in the tests do something like this (which would work in this test that I'm commenting on):
```ts
const localId = getListLocalId('First', 'Numbers');
should(this.ctx.localId_context[localId]).not.be.undefined();
```
This would mean that if a translator change (or a change in the test data CQL file) results in changing all the localIds, we wouldn't have to do this tedious task of fixing all the numbers again. On the other hand, it's not as explicit as just putting in the literal localId. What do you think?

test/elm/query/query-test.ts
```diff
@@ -433,27 +433,27 @@ describe('AggregateQuery', () => {
       new Interval(new DateTime(1982, 3, 16, 15, 0), new DateTime(2013, 5, 23, 10, 0))
     ];
     (await this.expressionStartingAggregation.exec(this.ctx)).should.eql(ret);
-    should(this.ctx.localId_context[30]).not.be.undefined();
+    should(this.ctx.localId_context[245]).not.be.undefined();
```
cmoesel, line 436: More localIds that are specified as numbers even though `localId_context` keys are technically strings (and so are localIds). And another place we could maybe just make a helper function to grab the localId from the data so we don't need to keep updating these tests.

test/spec-tests/skip-list.txt
```diff
@@ -5,15 +5,6 @@ CqlTypesTest.Time.TimeUpperBoundMinutes             Translation Error: Invalid t
...
-CqlLogicalOperatorsTest.Implies.TrueImpliesNull     Translation Error: Could not resolve call to operator Implies with signature (System.Boolean,System.Any).
-CqlLogicalOperatorsTest.Implies.FalseImpliesTrue    Translation Error: Could not resolve call to operator Implies with signature (System.Boolean,System.Boolean).
-CqlLogicalOperatorsTest.Implies.FalseImpliesFalse   Translation Error: Could not resolve call to operator Implies with signature (System.Boolean,System.Boolean).
-CqlLogicalOperatorsTest.Implies.FalseImpliesNull    Translation Error: Could not resolve call to operator Implies with signature (System.Boolean,System.Any).
-CqlLogicalOperatorsTest.Implies.NullImpliesTrue     Translation Error: Could not resolve call to operator Implies with signature (System.Any,System.Boolean).
-CqlLogicalOperatorsTest.Implies.NullImpliesFalse    Translation Error: Could not resolve call to operator Implies with signature (System.Any,System.Boolean).
-CqlLogicalOperatorsTest.Implies.NullImpliesNull     Translation Error: Could not resolve call to operator Implies with signature (System.Any,System.Any).
```
cmoesel, line 16: Fantastic! I love to see all these skip directives removed!

cmoesel (commented):
This is great stuff, @hossenlopp. Thank you!

I've left some comments about the tests that expect the context localId entries to be populated. My comments suggest fixing something that doesn't cause problems, but is technically incorrect (and was there previously, so _not your fault_) -- and an idea about how we can keep test maintenance easier in the future.

Since this was a previously existing issue, and you did not introduce any regression, I'm not "requesting changes", because it's valid for you to tell me it's out of scope for this PR and should be handled as a separate issue. But if you have the time and want to tackle it now rather than later, that's good too.

p9g:
https://github.com/cqframework/cql-execution/projects/1#card-52575839 move to done after merge

test/elm/list/list-test.ts
hossenlopp: I ended up getting a bit more generic with this since it wasn't always the first List item that was being looked for. I made a helper function that would traverse the ELM given a path to traverse and return the localId if it was successful.

test/elm/list/list-test.ts
```diff
@@ -671,37 +672,51 @@ describe('First', () => {
 
   it('should get first of a list of numbers and the localId should exist on the root context', async function () {
     (await this.numbers.exec(this.ctx)).should.equal(1);
-    should(this.ctx.localId_context[6]).not.be.undefined();
+    const listLocalId = getLocalIdByPath(data, 'First', 'Numbers', 'source');
+    should(listLocalId).not.be.null();
```
cmoesel, line 676: This is not harmful, but is it necessary? I assume that if `listLocalId` is `null`, then won't the test still fail on the next line anyway?

That said, I suppose it wouldn't be immediately obvious that it failed because it couldn't find the localId. But you could just have `getLocalIdByPath` throw an error if it can't find the id -- then the test would fail in a more obvious way when that happens.

Anyway... just a thought to possibly avoid having all these additional assertions.

test/elm/list/list-test.ts
cmoesel: Ah, good call!

cmoesel (approved):
Nice job, @hossenlopp -- and thanks for implementing that test helper for getting local ids. It was probably a little tedious up front, but it will probably save us a lot of tediousness going forward!

I left one comment for something to consider, but honestly, it's totally optional -- this is really good as it is. So... I'm approving. Thanks again for this great work!

test/elm/list/list-test.ts
hossenlopp: Good call! Implemented in [f0affad](https://github.com/cqframework/cql-execution/pull/320/commits/f0affadfd60e7b2892e34047f88a6b0fd6107ae8) and cleaned up the `not.be.null()` checks.

cmoesel (approved):
Perfect. Thank you, @hossenlopp!

dczulada (approved):

PR #326, opened 2024-10-11
Cache function ref

rdingwell:
This is primarily a performance enhancement.  When function use is defined within a Query, every iteration of elements in the where clause caused a re-evaluation of finding the correct FunctionDef to use.  When there are a large number of items to go through this becomes a very expensive process, mainly due to the type comparisons that are used to perform the FunctionDef lookups. 

This PR implements caching of the FunctionDef look ups with a FunctionRef the first time it is executed. 

Note that this PR also contains updates to 3 node modules that were not passing audit tests 

This is primarily a performance enhancement.  When function use is defined within a Query, every iteration of elements in the where clause caused a re-evaluation of finding the correct FunctionDef to use.  When there are a large number of items to go through this becomes a very expensive process, mainly due to the type comparisons that are used to perform the FunctionDef lookups. 

This PR implements caching of the FunctionDef look ups with a FunctionRef the first time it is executed. 

Note that this PR also contains updates to 3 node modules that were not passing audit tests 

**Submitter:**

- [] Tests are included and test edge cases

**Reviewer:**

Name:

### DIFF
--- a/examples/browser/cql4browsers.js
+++ b/examples/browser/cql4browsers.js
@@ -6811,21 +6811,21 @@ class FunctionRef extends expression_1.Expression {
         super(json);
         this.name = json.name;
         this.library = json.libraryName;
+        this.functionDefs = null;
     }
-    async exec(ctx) {
-        let functionDefs, child_ctx;
+    getFunctionDefs(ctx, args) {
+        if (this.functionDefs != null) {
+            // cache hit
+            return this.functionDefs;
+        }
+        let functionDefs;
         if (this.library) {
             const lib = ctx.get(this.library);
             functionDefs = lib ? lib.getFunction(this.name) : undefined;
-            const libCtx = ctx.getLibraryContext(this.library);
-            child_ctx = libCtx ? libCtx.childContext() : undefined;
         }
         else {
             functionDefs = ctx.get(this.name);
-            child_ctx = ctx.childContext();
         }
-        const args = await this.execArgs(ctx);
-        // Filter out functions w/ wrong number of arguments.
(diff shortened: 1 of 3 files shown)


cmoesel (approved):
My only concern with this approach was caching the resolved function for reuse across iterations, but I can't think of a use case where different iterations of the same FunctionRef would result in different function resolutions. I tried to force it with CQL like this:
```
define function ValueAsString(value FHIR.CodeableConcept):
  Coalesce(value.text.value, value.coding[0].display.value, value.coding[0].code.value, 'unknown')

define function ValueAsString(value FHIR.Quantity):
  ToString(value.value.value) + ' ' + Coalesce(value.unit.value, value.code.value)

define PatienObservationsAsStrings:
  [Observation] O return ValueAsString(O.value as Choice<FHIR.CodeableConcept, FHIR.Quantity>)
```
but the CQL-to-ELM compiler doesn't allow it because the function invocation is ambiguous. You have to do something like this, which results in two independency FunctionRefs:
```
define PatienObservationsAsStrings:
  [Observation] O return
    case
      when O.value is FHIR.CodeableConcept then ValueAsString(O.value as FHIR.CodeableConcept)
      when O.value is FHIR.Quantity then ValueAsString(O.value as FHIR.Quantity)
      else 'unsupported'
    end
```

All that to say, I _think_ this is safe and it appears to work. Still, I'd like for the FLAME team to have a chance to review and possibly regress it against their measures. @hossenlopp - do you agree that your team should review and regress this?

cmoesel:
I'll also note that the npm audit is unresolvable because the vulnerable module has not yet released a patch.

PR #329, opened 2024-12-04
ExpressionRef Library context_values performance issue

AymericGrassart:
A performance slowdown has been identified when executing referenced CQL statements under certain conditions.

### Assumptions
- Execution results of `ExpressionDef` expressions are cached in their library context. 
- `ExpressionRef` expressions use those cached results.

### Issue
The CQL below performs in a degraded, non-cached way. We suspect the wrong context is used to store or lookup the cached context values when ExpressionRefs cross library contexts.

### Test

CommonLib3:
```cql
library Common3
using Simple version '1.0.0'
context Patient

define function ExpensiveFunction():
  First(expand Interval[1, 100000])

define ExpensiveStatement:
  ExpensiveFunction()

define ExpensiveStatementRef:
  ExpensiveStatement union 
  ExpensiveStatement union 
  ExpensiveStatement union 
  ExpensiveStatement union 
  ExpensiveStatement
```

Using CommonLib3:
```
using Simple version '1.0.0'
include Common3 called common3
context Patient

define ExpensiveStatementRef: common3.ExpensiveStatementRef
```

The result of the library-test shows the perf degradation as if `ExpensiveStatement` was _not_ being cached:
```bash
$ npm test

  CommonLib3
    ✔ should be able to execute an expensive expression (123ms)

  Using CommonLib3
    ✔ should execute expression from included library at similar cost (565ms)
    ✔ should execute expression directly from included library at similar cost (291ms)
```

### Solution
?

Please note there are no changes to src code, only tests were added to demonstrate the perf issue. Looking for feedback.

We hope that a fix can be identified and included with this PR.
  
  
  
**Submitter:**

**Reviewer:**

Name:

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,7 +111,7 @@
     "build:spec-test-data": "node \"test/spec-tests/generate-cql.js\" && cd \"./test/generator\" && \"./gradlew\" generateSpecTestData && cd ..",
     "build:all": "npm run build && npm run build:browserify && npm run build:test-data && npm run build:spec-test-data",
     "watch:test-data": "cd \"./test/generator\" && \"./gradlew\" watchTestData && cd ..",
-    "test": "cross-env TS_NODE_PROJECT=\"./test/tsconfig.json\" TS_NODE_FILES=true mocha --reporter spec --recursive",
+    "test": "cross-env TS_NODE_PROJECT=\"./test/tsconfig.json\" TS_NODE_FILES=true mocha --reporter spec --recursive --grep CommonLib3",
     "test:nyc": "cross-env TS_NODE_PROJECT=\"./test/tsconfig.json\" TS_NODE_FILES=true nyc --reporter=html npx mocha --reporter spec --recursive",
     "test:watch": "cross-env TS_NODE_PROJECT=\"./test/tsconfig.json\" TS_NODE_FILES=true mocha --reporter spec --recursive --watch",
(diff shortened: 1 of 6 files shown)


rdingwell:
Hi @AymericGrassart, this issue has been addressed in #336.  Can you give us some additional context or use cases or test cases where you discovered the issue? The example certainly highlights the problem but is not something that seems all that practical in reality so we are curious as to how you discovered it.

AymericGrassart:
Hi @cmoesel and @rdingwell,
We really appreciate you digging into this issue and finding a root cause!
I realize that the repro example is not realistic, it was only created to show the underlying problem. For background, we have been using CQL for CDS for a few years and our clinical authors have created some pretty advanced logic for their programs. As the complexity increased, we used linked libraries to reference common logic for reuse. These libraries not only contained (uncached) functions but also statements, some with terminology-based retrieves. As the statement dependency tree grew, that's when we noticed the performance slowdown. It seemed to be most prevalent with subsequent ExpressionRef calls after crossing library boundaries. The asynchronous argument execution order resolution makes sense to ensure the same compute doesn't repeat and cached results are used.

We will we test the branch and let you know. 

For planning reasons, when do you expect to release a new 3.1.1 version for production use?

Thanks again!

hossenlopp:
This issue and fix has been cleaned up and merged in #336

Issue #324, opened 2024-09-05
Custom MessageListener not used when a CQL Message is in a referenced library


birick1:
Hi cql-execution team!

When implementing a [MessageListener](https://github.com/cqframework/cql-execution/blob/master/OVERVIEW.md#messagelistener), the listener works as expected when `Message(...)` is in the primary CQL file.  However, when `Message()` is in a referenced library, it does not work as expected.

For example, FHIRHelpers has several different `Message()` calls, but they will not trigger a call to an implemented `MessageListener`.

In debugging, a `NullMessageListener` would be instantiated for such library `Message` calls.  I discussed with @cmoesel, who identified that `this.messageListener` should be added as a final parameter in the constructors [here](https://github.com/cqframework/cql-execution/blob/master/src/runtime/context.ts#L440-L446) and [here](https://github.com/cqframework/cql-execution/blob/master/src/runtime/context.ts#L453-L459).  I tried out this solution, and it appears to have resolved the issue.

I'd like to ask that this solution be considered and added.

hossenlopp:
This issue was resolved in the [v3.2.0](https://github.com/cqframework/cql-execution/releases/tag/v3.2.0) release of cql-execution.

PR #330, opened 2025-01-16
Pass Message Listener to Included Libraries

nkarip:
Fix for cqframework/cql-execution#324

**Submitter:**

- [x] All dependent libraries are appropriately updated or have a corresponding PR related to this change - N/A

**Reviewer:**

Name:

### DIFF
--- a/test/elm/message/Included.js
+++ b/test/elm/message/Included.js
@@ -1,0 +1,10 @@
+/*
+   WARNING: This is a GENERATED file.  Do not manually edit!
+
+   To generate this file:
+       - Edit Included.cql to add a CQL Snippet
+       - From java dir: ./gradlew :cql-to-elm:generateTestData
+*/
+
+/* eslint-disable */
+
--- a/test/elm/message/Included.js
+++ b/test/elm/message/Included.js
@@ -7,4 +7,3 @@
 */
 
 /* eslint-disable */
-
--- a/test/elm/message/Included.js
+++ b/test/elm/message/Included.js
@@ -7,3 +7,4 @@
 */
 
 /* eslint-disable */
+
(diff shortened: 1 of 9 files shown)


test/elm/message/Included.js
```diff
@@ -0,0 +1,9 @@
+/*
+   WARNING: This is a GENERATED file.  Do not manually edit!
+
+   To generate this file:
+       - Edit Included.cql to add a CQL Snippet
+       - From java dir: ./gradlew :cql-to-elm:generateTestData
+*/
+
+/* eslint-disable */
```
cmoesel, line 9: This is admittedly a weird artifact of the test data build process, but... when I run `npm run build test-data`, this file shows changes because it has two more blank lines. I'm guessing that _prettier_ probably complained about them, so you tossed them. But... we don't want to have to do that every time the test data is regenerated, so it would be better to add this file to `.prettierignore` in the root of the project (you'll see other files like it in there) and check it in exactly the way it is generated.

Note: Although these changes should be checked in for this file, please continue to discard the generated changes in test/elm/type/data.js since those are due to a translator bug.

cmoesel (changes requested):
Thanks for fixing this. I'm surprised this mistake lasted so long in the codebase, but I guess people aren't using `Message` a lot in included libraries.

We run `npm audit` as part of our CI and it looks like it found a few issues. I don't think you introduced these, but would you mind trying to address them? Doing this on PRs ensures we keep dependencies clean before they get _too_ out of hand. I tried it locally and it should be as simple as running `npm audit fix`.

I also made one other request in a specific comment on the code.

Thanks!

cmoesel (approved):
This looks great. Thank you, Neelima!

hossenlopp (approved):
Looks good from the fqm-execution side of things. Good tests added too.

PR #332, opened 2025-02-13
Added Support for the ExpandValueset operator

sorlioglu:
**Description**

Added support for the [ExpandValueset](https://cql.hl7.org/09-b-cqlreference.html#expandvalueset) operator by adding the [expand](https://github.com/sorliog/cql-execution/blob/ef9ff9d694f4c4c2ecb5c934c329a9298ca9d7c5/src/datatypes/clinical.ts#L94) method in the Valueset class. The [expand](https://github.com/sorliog/cql-execution/blob/ef9ff9d694f4c4c2ecb5c934c329a9298ca9d7c5/src/datatypes/clinical.ts#L96) method is invoked in [ExpandValueSet.exec()](https://github.com/sorliog/cql-execution/blob/ef9ff9d694f4c4c2ecb5c934c329a9298ca9d7c5/src/elm/clinical.ts#L111) expression. 

**Submitter:**

**Reviewer:**

Name:

### DIFF
--- a/src/cql-code-service.ts
+++ b/src/cql-code-service.ts
@@ -1,5 +1,5 @@
 import {Code, CodeSystem, ValueSet, Concept} from './datatypes/datatypes';
-import {AdvancedTerminologyProvider, TerminologyProvider, ValueSetDictionary, ValueSetObject} from './types';
+import { TerminologyProvider, ValueSetDictionary, ValueSetObject} from './types';
 
 class CodeService implements TerminologyProvider {
   valueSets: ValueSetObject;
@@ -44,85 +44,104 @@ class CodeService implements TerminologyProvider {
 }
 
 
-class AdvancedCodeService implements AdvancedTerminologyProvider {
-  valueSets: ValueSetObject;
+// class AdvancedCodeService implements AdvancedTerminologyProvider {
+//   valueSets: ValueSetObject;
+//
+//   constructor(valueSetsJson: ValueSetDictionary = {}) {
+//     this.valueSets = {};
+//     for (const oid in valueSetsJson) {
+//       this.valueSets[oid] = {};
+//       for (const version in valueSetsJson[oid]) {
+//         const codes = valueSetsJson[oid][version].map(
+//           (code: any) => new Code(code.code, code.system, code.version)
+//         );
+//         this.valueSets[oid][version] = new ValueSet(oid, version, codes);
+//       }
+//     }
+//   }
+//
+//   findValueSetsByOid(oid: string): ValueSet[] {
+//     return this.valueSets[oid] ? Object.values(this.valueSets[oid]) : [];
+//   }
+//
+//   findValueSet(oid: string, version?: string): ValueSet | null {
+//     if (version != null) {
+//       return this.valueSets[oid] != null ? this.valueSets[oid][version] : null;
+//     } else {
+//       const results = this.findValueSetsByOid(oid);
+//       if (results.length === 0) {
+//         return null;
+//       } else {
+//         return results.reduce((a: any, b: any) => {
+//           if (a.version > b.version) {
+//             return a;
+//           } else {
+//             return b;
+//           }
+//         });
+//       }
+//     }
+//   }
+//
+//
+//   expandValueSet(oid: string, version?: string): Code[] {
+//     const results: Code[] = [];
+//     const valueSet: ValueSet | null = this.findValueSet(oid, version);
+//     if (valueSet == null) {
+//       return [];
+//     }
+//     valueSet.codes.forEach(code => {
+//       if (!results.find(result => {
+//         result === code;
+//       })) {
+//         results.push(code);
+//
+//       }
+//     });
+//     return results;
+//   }
+//
+//
+//   inValueSet(code: Code | Code[], oid: string, version?: string): boolean {
+//     const valueSet: ValueSet | null = this.findValueSet(oid, version);
+//     if (valueSet == null) {
+//       return false;
+//     }
+//
+//     const codeList: Code[] = Array.isArray(code) ? code : [code];
+//
+//     return codeList.every(singleCode => {
+//       valueSet.hasMatch(singleCode);
+//     });
+//   }
+//
+//   anyInValueSet(code: Code | Code[], oid: string, version?: string): boolean {
+//     const valueSet: ValueSet | null = this.findValueSet(oid, version);
+//     if (valueSet == null) {
+//       return false;
+//     }
+//
+//     const codeList: Code[] = Array.isArray(code) ? code : [code];
+//
+//     return codeList.some(singleCode => {
+//       valueSet.hasMatch(singleCode);
+//     });
+//   }
+//
+//
+//   expandCodeSystem(codeSystem: CodeSystem): Code[] {
+//     throw new Error('This function is not implemented.');
+//   }
+//
+//
+//   inCodeSystem(code: Code | Code[], codeSystem: CodeSystem): boolean {
+//     throw new Error('This function is not implemented.');
+//   }
+//
+//   subsumes(subsuming: Code | Concept, subsumed: Code | Concept): boolean {
+//     throw new Error('This function is not implemented.');
+//   }
+// }
 
-  constructor(valueSetsJson: ValueSetDictionary = {}) {
-    this.valueSets = {};
-    for (const oid in valueSetsJson) {
-      this.valueSets[oid] = {};
-      for (const version in valueSetsJson[oid]) {
-        const codes = valueSetsJson[oid][version].map(
-          (code: any) => new Code(code.code, code.system, code.version)
-        );
-        this.valueSets[oid][version] = new ValueSet(oid, version, codes);
-      }
-    }
-  }
 
-  findValueSetsByOid(oid: string): ValueSet[] {
-    return this.valueSets[oid] ? Object.values(this.valueSets[oid]) : [];
-  }
-
-  findValueSet(oid: string, version?: string): ValueSet | null {
-    if (version != null) {
-      return this.valueSets[oid] != null ? this.valueSets[oid][version] : null;
-    } else {
-      const results = this.findValueSetsByOid(oid);
-      if (results.length === 0) {
-        return null;
-      } else {
-        return results.reduce((a: any, b: any) => {
-          if (a.version > b.version) {
-            return a;
-          } else {
-            return b;
-          }
-        });
-      }
-    }
-  }
-
-
-
-  expandValueSet(oid: string, version?: string): Code[] {
-    const results: Code[] = [];
-    const valueSet: ValueSet | null = this.findValueSet(oid, version);
-    if (valueSet == null) {
-      return [];
-    }
-    valueSet.codes.forEach(code => {
-      if (!results.find(result => {
-        result === code;
-      })) {
-        results.push(code);
-
-      }
-    });
-    return results;
-  }
-
-
-
-  inValueSet(code: Code | Code[], oid: string, version?: string): boolean {
-    throw new Error('This function is not implemented.');
-  }
-
-
-  expandCodeSystem(codeSystem: CodeSystem): Code[] {
-    throw new Error('This function is not implemented.');
-  }
-
-
-  inCodeSystem(code: Code | Code[], codeSystem: CodeSystem): boolean {
-    throw new Error('This function is not implemented.');
-  }
-
-  subsumes(subsuming: Code | Concept, subsumed: Code | Concept): boolean {
-    throw new Error('This function is not implemented.');
-  }
-}
-
-
-
-export {CodeService,AdvancedCodeService};
+export {CodeService};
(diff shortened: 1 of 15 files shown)


src/cql-code-service.ts
```diff
@@ -40,3 +40,104 @@ export class CodeService implements TerminologyProvider {
...
+//
+//
+//   inCodeSystem(code: Code | Code[], codeSystem: CodeSystem): boolean {
+//     throw new Error('This function is not implemented.');
+//   }
+//
+//   subsumes(subsuming: Code | Concept, subsumed: Code | Concept): boolean {
+//     throw new Error('This function is not implemented.');
+//   }
+// }
```
cmoesel, line 141: Generally speaking, we try to avoid having large blocks of commented out code. If this code is not needed, I'd prefer that we delete it.

src/cql-code-service.ts
```diff
@@ -1,7 +1,7 @@
 import { Code, ValueSet } from './datatypes/datatypes';
 import { TerminologyProvider, ValueSetDictionary, ValueSetObject } from './types';
 
-export class CodeService implements TerminologyProvider {
+class CodeService implements TerminologyProvider {
```
cmoesel, line 4: In order to remain consistent with the rest of the code base, I think it's best to export the class declaration here (as it was before).

src/cql-code-service.ts
```diff
@@ -40,3 +40,104 @@ export class CodeService implements TerminologyProvider {
...
+//   inCodeSystem(code: Code | Code[], codeSystem: CodeSystem): boolean {
+//     throw new Error('This function is not implemented.');
+//   }
+//
+//   subsumes(subsuming: Code | Concept, subsumed: Code | Concept): boolean {
+//     throw new Error('This function is not implemented.');
+//   }
+// }
+
+export { CodeService };
```
cmoesel, line 143: As noted above, in order to remain consistent with the rest of the code base, we should export as part of the class declaration.

src/elm/expressions.ts
```diff
@@ -24,6 +24,8 @@ export * from './structured';
 export * from './type';
 export * from './overloaded';
 
+// TODO Add advanced valueset operations here
+
```
cmoesel, line 28: I think you can delete this.

src/types/cql-code-service.interfaces.ts
```diff
@@ -30,3 +30,17 @@ export interface TerminologyProvider {
...
+//
+// export interface AdvancedTerminologyProvider extends TerminologyProvider {
+//   inValueSet: (code: Code | Code[], oid: string, version?: string) => boolean | Promise<boolean>;
+//   anyInValueSet: (code: Code | Code[], oid: string, version?: string) => boolean | Promise<boolean>;
+//   expandValueSet: ( oid: string, version?: string) => Code[] | Promise<Code[]>;
+//   inCodeSystem: (code: Code | Code[], codeSystem: CodeSystem) => boolean | Promise<boolean>;
+//   // expandCodeSystem: (codeSystem: CodeSystem) => Code[] | Promise<Code[]>;
+//   // subsumes: (subsuming: Code | Concept, subsumed: Code | Concept) => boolean | Promise<boolean>;
+//
+// }
```
cmoesel, line 46: This should be deleted as well.

cmoesel (changes requested):
Thank you for this contribution, @sorliog! This looks pretty good to me. I left a few comments regarding small things to change -- but all the important parts seems to be working well!

We like to have more than one review of each PR, so I'm also going to ask another colleague to review this.

Thanks again! We love receiving contributions!

cmoesel:
I've also just realized that while you added datatype tests, there are no tests for the ELM ExpandValueSet expression. Could you add some cql tests in `test/elm/clinical` to test expansion of value sets? You could either use the [ExpandValueSet](https://cql.hl7.org/09-b-cqlreference.html#expandvalueset) CQL operator (if the CQL-to-ELM translator supports it) or you could do a union of value sets to force expansion. There is some information about developing this type of test [here](https://github.com/cqframework/cql-execution?tab=readme-ov-file#to-develop-tests), but you can also just look at the existing tests to get a feel for it.

There is one current _gotcha_. When you run `npm run build:test-data`, it will generate some invalid ELM changes in the `test/elm/type/data.js` file due to a bug in CQL-to-ELM. You need to revert those changes for the tests to run (but obviously, keep the changes in `test/elm/clinical/data.js`.

sorliog:
> I've also just realized that while you added datatype tests, there are no tests for the ELM ExpandValueSet expression. Could you add some cql tests in `test/elm/clinical` to test expansion of value sets? You could either use the [ExpandValueSet](https://cql.hl7.org/09-b-cqlreference.html#expandvalueset) CQL operator (if the CQL-to-ELM translator supports it) or you could do a union of value sets to force expansion. There is some information about developing this type of test [here](https://github.com/cqframework/cql-execution?tab=readme-ov-file#to-develop-tests), but you can also just look at the existing tests to get a feel for it.
> 
> There is one current _gotcha_. When you run `npm run build:test-data`, it will generate some invalid ELM changes in the `test/elm/type/data.js` file due to a bug in CQL-to-ELM. You need to revert those changes for the tests to run (but obviously, keep the changes in `test/elm/clinical/data.js`.

Understood. I made the requested changes. Let me know if I missed anything.

src/elm/expressions.ts
cmoesel: It looks like this can still be deleted.

test/elm/clinical/clinical-test.ts
```diff
@@ -1,10 +1,14 @@
 import should from 'should';
 import setup from '../../setup';
+
 const vsets = require('./valuesets');
 import * as DT from '../../../src/datatypes/datatypes';
 import { Uncertainty } from '../../../src/datatypes/uncertainty';
+
 const { p1, p2, p3, p4, p5, p6, p7 } = require('./patients');
 import { PatientSource } from '../../../src/cql-patient';
+import { Code } from '../../../lib/datatypes/clinical';
```
cmoesel, line 10:
```suggestion
import { Code } from '../../../src/datatypes/clinical';
```

test/elm/clinical/clinical-test.ts
```diff
@@ -157,6 +161,48 @@ describe('InValueSet', () => {
...
+
+  it('expand valueset', async function () {
+    const received: any = await this.expandFemale.exec(this.ctx);
+    const expected: Code[] = [new Code('F', '2.16.840.1.113883.18.2', 'HL7V2.5', undefined)];
+    (received.length == expected.length).should.true();
+    (received[0] != null).should.true();
+    (received[0].code === expected[0].code).should.true();
+    (received[0].system === expected[0].system).should.true();
+    (received[0].version === expected[0].version).should.true();
+    (received[0].display === expected[0].display).should.true();
```
cmoesel, line 177: This can be simplified to:
```ts
received.should.eql(expected);
```
This didn't work initially because it was importing `Code` from the `lib` folder instead of the `src` folder. But once you fix that import, `eql` works as expected here.

test/elm/clinical/clinical-test.ts
```diff
@@ -157,6 +161,48 @@ describe('InValueSet', () => {
...
+    ];
+    (received.length == expected.length).should.true();
+
+    for (let i = 0; i < received.length; i++) {
+      (received[i] != null).should.true();
+      (received[i].code === expected[i].code).should.true();
+      (received[i].system === expected[i].system).should.true();
+      (received[i].version === expected[i].version).should.true();
+      (received[i].display === expected[i].display).should.true();
+    }
```
cmoesel, line 202: This can also be simplified to:
```ts
received.should.eql(expected);
```

cmoesel (changes requested):
Hello again. I apologize for the delay... This is really close! I just made a couple more comments:
- A now-irrelevant code comment that hasn't been deleted yet
- A couple of small tweaks in the tests

sorliog:
> Hello again. I apologize for the delay... This is really close! I just made a couple more comments:
> 
> * A now-irrelevant code comment that hasn't been deleted yet
> * A couple of small tweaks in the tests

No problem. I made the requested changes. One thing to note is I had to run `npm audit fix`; it upgraded `node_modules/elliptic` to version 6.6.1 due to a vulnerability in the previous version.

cmoesel (approved):
Thank you for the contribution, @sorliog. This is a nice feature to have.

Everything looks good to me. We like to have two reviews on each PR, so I'm going to ask someone else to look at it as well.

In the mean time, could you either rebase or merge master so your branch is up to date? We just merged in another PR this morning.

test/elm/clinical/clinical-test.ts
```diff
@@ -157,6 +161,35 @@ describe('InValueSet', () => {
...
+  it('expand valueset', async function () {
+    const received: any = await this.expandFemale.exec(this.ctx);
+    const expected: Code[] = [new Code('F', '2.16.840.1.113883.18.2', 'HL7V2.5', undefined)];
+    received.should.eql(expected);
+  });
+
+  it('expand null', async function () {
+    this.expandNull
+      .exec(this.ctx)
+      .should.be.rejectedWith('ValueSet must be provided to ExpandValueSet function');
```
hossenlopp, line 178: This should expect a `null` result as defined in the CQL specification. https://cql.hl7.org/04-logicalspecification.html#expandvalueset and https://cql.hl7.org/09-b-cqlreference.html#expandvalueset

src/elm/clinical.ts
```diff
@@ -99,6 +100,24 @@ export class InValueSet extends Expression {
...
+
+  constructor(json: any) {
+    super(json);
+    this.valueset = new ValueSetRef(json.operand);
+  }
+
+  async exec(ctx: Context) {
+    const valueset = await this.valueset.execute(ctx);
+    if (valueset == null || !valueset.isValueSet) {
+      throw new Error('ValueSet must be provided to ExpandValueSet function');
```
hossenlopp, line 114: This should return a `null` result if the operand result is `null`. If it is not a ValueSet, then it would still make sense for the error to be thrown.

This is defined in the CQL specification. https://cql.hl7.org/04-logicalspecification.html#expandvalueset and https://cql.hl7.org/09-b-cqlreference.html#expandvalueset

hossenlopp (changes requested):
ExpandValueSet should return `null` if the operand is `null`.

src/elm/clinical.ts
```diff
@@ -99,6 +100,24 @@ export class InValueSet extends Expression {
...
+
+  constructor(json: any) {
+    super(json);
+    this.valueset = new ValueSetRef(json.operand);
+  }
+
+  async exec(ctx: Context) {
+    const valueset = await this.valueset.execute(ctx);
+    if (valueset == null || !valueset.isValueSet) {
+      return null;
```
hossenlopp, line 114: I think it should still throw the error if it isn't a ValueSet but not null. I think this situation could occur if there is something wonky with the terminology provider.

hossenlopp (changes requested):

src/elm/clinical.ts
cmoesel: @sorliog -- are you able to make this last change that @hossenlopp has requested?

sorliog: @cmoesel @hossenlopp I did make the change. See [commit 3469b74](https://github.com/cqframework/cql-execution/pull/332/commits/3469b7443fc46d7138279f455f8a27ed386045e2)

cmoesel: @sorliog - That commit returns null for a null value set _and_ an object that is not a value set at all. @hossenlopp's suggestion is that if we have a non-null value, but it's _not_ a ValueSet, then it is still appropriate to throw an error in that case. So something more like:
```ts
if (valueset == null) {
  return null;
} else if (!valueset.isValueSet) {
  throw new Error('ExpandValueSet function invoked on object that is not a ValueSet');
}
```

sorliog: @cmoesel I understand. I can make the change. However, I don't see a way to unit test whether the exception is thrown because the [elm/library.ts](https://github.com/sorliog/cql-execution/blob/3469b7443fc46d7138279f455f8a27ed386045e2/src/elm/library.ts#L106) evaluates non-valuesets to `undefined`. Therefore ` if (!valueset.isValueSet)` will never be triggered, since anything that is not a valueset is returned as an `undefined`.

cmoesel: I agree. It would be difficult to test. I think that if we wanted to test it, we'd need to override the CodeService in the test's executor in order to make it return a non-ValueSet where a ValueSet is expected. IMO, the ROI on such a test is not very high -- so I think I'd be OK with that branch of code (the one that throws on a non-VS) being uncovered. @hossenlopp - do you agree that it's probably ok not to have a test for the edge case where a noncompliant CodeService is used?

sorliog: @cmoesel Changes committed See [this commit](https://github.com/cqframework/cql-execution/pull/332/commits/6cc6f1ec4307e408165c39cb805b2b88e5aac74b)

hossenlopp: This makes sense! We can't coerce the translator to make something like this:
```
define ExpandString: ExpandValueSet('string' as ValueSet)
```
I tried.... It nulled out the entire thing. So we can't easily make a test for it, and that's okay.

hossenlopp (approved):
Looks good to me!

cmoesel (approved):
Yes! This looks great. Thank you so much, @sorliog!

PR #342, opened 2025-03-31
Disable console colors in tests

cmoesel:
Later versions of Node 20 output errors in red text. This causes one of our tests to fail because it does not expect color codes. This PR adds `NO_COLOR=1` to the test calls to ensure that the tests are run without consle colors.

There is an alternate PR (#343) that solves this issue another way: it simply makes the test more forgiving (and ignores the color control codes). I think I prefer the other PR because it will work when you run the tests via other methods (like an IDE plugin) as well -- whereas this only fixes it for tests invoked through the NPM scripts.

**Submitter:**

**Reviewer:**

Name:


cmoesel:
Closing in favor of #343.

PR #343, opened 2025-03-31
Ignore console collor in message listener test

cmoesel:
One of the message listener tests fails on some versions of Node 20 because it is not expecting console output to use control characters for color. This PR makes the test look only for the expected message and ignores any other characters.

There is an alternate PR (#342) that fixes the same problem by setting the `NO_COLOR` environment variable in the NPM test scripts. I think I prefer this current PR, however, since it should work no matter how you invoke the tests (for example, via an IDE plugin).

**Submitter:**

**Reviewer:**

Name:


hossenlopp (approved):
Trivial change that make sense! Other tests that look for errors in Expression work rely on regex matching since errors get wrapped with extra info on the ELM location.

Issue #226, opened 2021-02-18
Support ValueSet references as arguments to functions


cmoesel:
See: https://chat.fhir.org/#narrow/stream/179220-cql/topic/ValueSet.20Typing

JSRankins:
Hi @cmoesel. We were discussing ValueSetRef and CodeSystemRef as first-class types at the Connecathon today. An update is being made to fix some issues with  ValueSetRef and CodeSystemRef as first-class types in CQL with version 1.5.2. 1.5.2 correlates to [CQL-to-ELM Translator 1.5.6](https://github.com/cqframework/clinical_quality_language/releases/tag/v1.5.6). Bryn had mentioned this functionality wasn't available in cql-execution yet. He said a ticket needed to be created, so I volunteered. See [Issue 259](https://github.com/cqframework/cql-execution/issues/259). I didn't see this ticket until afterward. Not sure if you want to merge tickets or, since this ticket seems to be specific to ValueSet references, leave Issue 259 open for the CodeSystem reference piece. Thoughts?

cmoesel:
I think it's likely we might implement these separately (one at a time), so it probably makes sense to have one issue for VS references and one for CS references.  @JSRankins - can you update #259 to be specific to CodeSystems? And also include a reference to this ticket so that at least they are linked in some way?

JSRankins:
@cmoesel Done.

IrickNcqa:
Adding a note of support to this request from NCQA.  A use case from our measures prompted this language update, and as such we are looking for engine support for the feature.

PR #349, opened 2025-08-12
Vs firstclass

lmd59:
Addresses https://github.com/cqframework/cql-execution/issues/226 by creating a new type (`CQLValueSet`) that aligns with the CQL 1.5 specification defined ValueSet type. 
Note: This type naming is currently `CQLValueSet` so as not to change the interface of `ValueSet`, but this is intended to be a temporary decision that should be updated for the next major update. At the next major update, change `CQLValueSet` -> `ValueSet` and `ValueSet` -> `ValueSetExpansion`.

This PR shifts valueset resolution to the point that it is required, such as when `hasMatch` or `expand` is called. It does not cache valueset expansions in the context and leaves valueset caching up to the code service. This includes an update to equivalence that only expands valuesets to check matching codes if the valuesets do not have an exact match of id/version. The PR also adds a consistent error if passed valuesets are not able to be resolved by the code service.

**Submitter:**

- [✔ ] This pull request describes why these changes were made
- [✔ ] Code diff has been done and been reviewed (it does not contain: additional white space, not applicable code changes, debug statements, etc.)
- [✔ ] Tests are included and test edge cases
- [✔ ] Tests have been run locally and pass
- [~ ] Code coverage has not gone down and all code touched or added is covered. 
- [✔ ] Code passes lint and prettier (hint: use `npm run test:plus` to run tests, lint, and prettier)
- [✔ ] All dependent libraries are appropriately updated or have a corresponding PR related to this change
- [✔ ] `cql4browsers.js` built with `npm run build:browserify` if source changed.

~ _Code coverage has not gone down, but not all added/touched code is covered as code lines represent unlikely cases that would be difficult to test._

**Reviewer:**

Name:


src/elm/clinical.ts
```diff
@@ -2,26 +2,34 @@ import { Expression } from './expression';
...
   async exec(ctx: Context) {
-    const valueset =
-      (await ctx.codeService.findValueSet(this.id, this.version)) ||
-      new dt.ValueSet(this.id, this.version);
-    ctx.rootContext().set(this.name, valueset);
+    let codeSystems;
+    if (this.codesystems) {
+      codeSystems = await Promise.all(
+        this.codesystems.map(async cs => {
+          const csdef = (await cs.exec(ctx)) as CodeSystemDef;
```
hossenlopp, line 26: I feel like we should just execute and return the CodeSystemDef here.

hossenlopp (changes requested):
First pass review. I mainly looked from the aspect of the change in code coverage. Found a few things worth changing and others worth discussing.

src/elm/clinical.ts
lmd59: This is used to create the `CQLValueSet`, which requires an array of `CodeSystem`s

hossenlopp: Yes. But executing the `CodeSystemDef`s would give us the `CodeSystem` objects.

lmd59: It looks like it should actually already be executed when the CodeSystemRef is executed. I will make that change!

src/elm/clinical.ts
```diff
@@ -148,7 +154,12 @@ export class CodeSystemRef extends Expression {
 
   async exec(ctx: Context) {
     const codeSystemDef = ctx.getCodeSystem(this.name, this.libraryName);
-    return codeSystemDef ? codeSystemDef.execute(ctx) : undefined;
+    if (!codeSystemDef) {
```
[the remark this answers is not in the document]
elsaperelli: Hmmm I was kind of going off of `CodeRef` and `ConceptRef` for this. Do we need to do that for them as well if this is the case? Is it using a codeservice like the valueset resolver is?

src/datatypes/clinical.ts
```diff
@@ -36,14 +36,35 @@ export class Concept {
...
+  constructor(
+    public id: string,
+    public version?: string,
+    public name?: string,
+    public codesystems?: CodeSystem[]
+  ) {
+    super(id, version, name);
   }
 
   get isValueSet() {
```
elsaperelli, line 59: Do we want this in the `ValueSet` class too?

test/datatypes/clinical-test.ts
```diff
@@ -119,6 +119,21 @@ describe('Concept', () => {
   });
 });
 
+describe('ValueSet', () => {
```
elsaperelli, line 122: Do we want to title this `CQLValueSet`?

test/elm/external/external-test.ts
```diff
@@ -1,8 +1,9 @@
 import should from 'should';
 import 'should-sinon';
 import sinon from 'sinon';
-import { Code, Interval, Repository, RetrieveDetails, ValueSet } from '../../../src/cql';
+import { Code, Interval, Repository, RetrieveDetails } from '../../../src/cql';
 import setup from '../../setup';
+import { ValueSet } from '../../../src/datatypes/clinical';
```
elsaperelli, line 6: In this file, you import `ValueSet` from `src/datatypes/clinical`, but in `test/util/immutableUtil-test.ts`, you just change the import from `ValueSet` to `CQLValueSet` from `src/cql`. Is this difference intentional?

test/util/immutableUtil-test.ts
```diff
@@ -1,6 +1,6 @@
 import Immutable from 'immutable';
 import should from 'should';
-import { Code, Concept, DateTime, Quantity, Ratio, ValueSet } from '../../src/cql';
+import { Code, CodeSystem, Concept, DateTime, Quantity, Ratio, CQLValueSet } from '../../src/cql';
```
elsaperelli, line 3: Same comment as `external-test.ts`

elsaperelli (changes requested):
Just a few comments/questions for now but looks good!

src/elm/clinical.ts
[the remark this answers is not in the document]
lmd59: It's not using the codeservice, it's bubbling up through context to the library object with the assumption that it should be loaded in there. But if we have a reference to a code system, shouldn't we expect it to be loaded?

test/datatypes/clinical-test.ts
lmd59: Rawr- this is what I get for doing a last minute symbol rename. lol- thanks for catching!

src/elm/clinical.ts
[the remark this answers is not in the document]
lmd59: Ok- I could go either way then. If we don't throw an error, I'd probably just go straight to `return codeSystemDef.execute(ctx);` without the codeSystemDef ternary though.

src/elm/external.ts
```diff
@@ -33,13 +33,20 @@ export class Retrieve extends Expression {
...
 
-      retrieveDetails.codes = resolvedCodes;
+      if (typeIsArray(executedCodes)) {
+        retrieveDetails.codes = executedCodes as Code[];
+      } else if (executedCodes) {
+        // retrieveDetails codes are expected to be expanded for external usage
+        retrieveDetails.codes = await resolveValueSet(executedCodes as CQLValueSet, ctx);
+      } else {
+        retrieveDetails.codes = undefined;
+      }
```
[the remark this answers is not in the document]
hossenlopp: The previous code would have entered the if statement and just returned `[]` when `resolvedCodes` was `null` or `undefined` because of the `==` comparison to `null`. Since that comparison is still there, it still leaves the function early and and will never make it to this else.

src/elm/clinical.ts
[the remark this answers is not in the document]
lmd59: Done!

hossenlopp (changes requested):
Got a few more suggestions on tests.

elsaperelli (approved):
Lgtm! 🥳

elsaperelli (approved):

hossenlopp (approved):

cmoesel (approved):
Fantastic work, @lmd59 -- and great reviews, @hossenlopp and @elsaperelli!

I left one tiny comment about a test that I think could be improved, but I wouldn't let it hold up this PR from being merged. So... if @lmd59 has time to address it, great! Let's do that! But if she doesn't has time, then LGTM!

test/elm/clinical/clinical-test.ts
```diff
@@ -16,23 +16,36 @@ describe('ValueSetDef', () => {
...
-    vs.oid.should.equal('1.2.3.4.5.6.7.8.9');
+    vs.id.should.equal('1.2.3.4.5.6.7.8.9');
+    vs.version.should.equal('1');
+  });
+
+  it('should execute three-arg to ValueSet with ID and version', async function () {
+    const vs = await this['unknown Three Arg'].exec(this.ctx);
+    vs.id.should.equal('1.2.3.4.5.6.7.8.9');
     vs.version.should.equal('1');
+    vs.codesystems.length.should.equal(1);
```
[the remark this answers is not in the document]
cmoesel: Perfect. Thank you!

PR #360, opened 2026-04-16
CQL Test Server for CQL Tests Runner

cmoesel:
This PR implements a test server that exposes a `$cql` target to allow the CQL Tests Runner to run tests against the `cql-execution` engine. Rather than repeat all the details here, I'll refer you to the [test-server/README.md](https://github.com/cqframework/cql-execution/blob/test-server/test-server/README.md) file in this branch.

I initially developed this as a standalone project outside of `cql-execution` that pulled in `cql-execution` as an NPM dependency. This PR integrates it into the `cql-execution` project as a sub-project with its own `package.json` and configuration files. In addition to general review of the code, I'm also interested in thoughts on this approach. Should we:
* Keep it as I have it here (a sub-project)? _UPDATE: We chose this approach._
* Keep it in `cql-execution` but try to integrate it more closely (i.e., share `package.json` and configuration files)?
* Or introduce it as a completely separate project outside of `cql-execution` (and if so, where)?

~I have not yet invested in updating GitHub actions to automatically run the tests or other checks in the sub-project. I will do that after we have confirmed the approach.~ _UPDATE: GitHub Actions are now in place._

_NOTE: The npm audit check is currently failing due to a reported vulnerability in elliptic. There is currently no fix for that._

**Submitter:**

- [n/a] `cql4browsers.js` built with `npm run build:browserify` if source changed.

**Reviewer:**

Name:


hossenlopp (commented):
This approach makes sense. The embedded project makes it always available for running.

I'm not 100% sure if it is universally using cql-execution in the parent directory when cql-exec-fhir is being used. Looking in the package-lock.json shows it is referencing the version on npm. This could cause some oddities with the PatientSource created by cql-exec-fhir, which may be using the npm version instead of the local.

I would try setting up cql-execution as a dependency to `../` and using cql-execution "normally", this hopefully will make sure cql-exec-fhir and your runner code are using the same cql-execution.

test-server/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+```sh
+npm run dev
+```
+
+The development mode builds the `cql-execution` source (to ensure it uses the latest changes) before running the server. It watches for changes to the `test-server` source and hot-reloads as necessary. By default, it uses the javascript CQL-to-ELM translator and listens on `http://localhost:8000`.
+
+_NOTE: The watch mode will NOT detect or hot-reload changes in the `cql-execution` source. You must stop and restart the `test-server` to pick up changes in the `cql-execution` source._
+
+# Production mode::
```
elsaperelli:
```suggestion
# Production mode:
```

test-server/README.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+The production mode builds the `cql-execution` source and `test-server` source then runs the server _without_ watching for changes. By default, it uses the javascript CQL-to-ELM translator and listens on `http://localhost:8000`.
+
+### Configuration
+
+- `PORT` (optional): Port to bind. Defaults to `8000`.
+- `USE_TRANSLATION_SERVICE` (optional, default: `false`): If `true`, the server will use a local [CQL Translation Service](https://github.com/cqframework/cql-translation-service) for CQL-to-ELM. In this case, the CQL Translation service must be running and available at `http://localhost:8080`. If `false`, the server will use the javascript CQL-to-ELM module to compile CQL.
+
+You can export these variable before running the server or place them in a `.env` file:
```
elsaperelli:
```suggestion
You can export these variables before running the server or place them in a `.env` file:
```

test-server/README.md
```diff
@@ -0,0 +1,115 @@
...
+  - Example request:
+    ```json
+    {
+      "resourceType": "Parameters",
+      "parameter": [{ "name": "expression", "valueString": "1 + 2" }]
+    }
+    ```
+  - Example response:
+    ```json
+    {
```
elsaperelli, line 17: Just curious- would it be worth including the cqf-cqlType extension in this example response?

test-server/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+```ini
+PORT=8000
+USE_TRANSLATION_SERVICE=false
+```
+
+## Quick Test With curl
+
+```sh
+curl -X POST http://localhost:8000/fhir/\$cql \
```
elsaperelli, line 93: Hm when I just copied and pasted this in my terminal I get the following error: `{"error":"Missing 'expression' parameter with CQL valueString in request body"}`

test-server/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+### Development mode:
+
+```sh
+npm run dev
+```
+
+The development mode builds the `cql-execution` source (to ensure it uses the latest changes) before running the server. It watches for changes to the `test-server` source and hot-reloads as necessary. By default, it uses the javascript CQL-to-ELM translator and listens on `http://localhost:8000`.
+
+_NOTE: The watch mode will NOT detect or hot-reload changes in the `cql-execution` source. You must stop and restart the `test-server` to pick up changes in the `cql-execution` source._
```
elsaperelli, line 67: Should the watch mode be able to detect changes in `test-server/src/operation.ts`?

test-server/src/operation.ts
```diff
@@ -0,0 +1,38 @@
+import { Parameters } from 'fhir/r4';
+import logger from './logger';
+import { translate } from './translate';
+import { CodeService, DateTime, Executor, Library } from '../..';
+import { PatientSource } from 'cql-exec-fhir';
+import { toParameters } from './convert';
+
+const USE_TRANSLATION_SERVICE = process.env['USE_TRANSLATION_SERVICE'] == 'true';
```
elsaperelli: I think this needs to be moved somewhere else, right now when I add a `.env` file and change `USE_TRANSLATION_SERVICE` to true, it still uses the javascript translator. I did some investigation and it looks like before `translate` `USE_TRANSLATION_SERVICE` is `false` even though `process.env['USE_TRANSLATION_SERVICE']` is true.

elsaperelli (changes requested):
This is awesome!!! Posting some of the review comments I have so far!

tsconfig.json
```diff
@@ -11,7 +11,7 @@
     "forceConsistentCasingInFileNames": true,
     "declaration": true,
     "sourceMap": true,
-    "moduleResolution": "node",
-    "module": "CommonJS"
+    "moduleResolution": "nodenext",
```
elsaperelli, line 15: Not sure how much this matters, but since this changed I am seeing an error in `src/runtime/messageListeners.ts` about not finding the name `console`. If you add `types: ["node"]` to these compilerOptions that should fix that!

test-server/README.md
cmoesel: Sure. I added the cqf-cqlType extension to everything toward the end (after I initially wrote the README), so that's probably why it's not in there. Good catch!

test-server/README.md
cmoesel: Good find. It turns out that the server was only accepting content-type `application/json`, so `application/fhir+json` was not being parsed correctly. 

Doh. I'm embarrassed that (a) I didn't catch that bug earlier, and (b) I apparently didn't test the curl command myself (which honestly surprises me).

test-server/README.md
cmoesel: Yes, it should. And it's odd that it's not doing that, because I swear I used this feature in development. In addition, the [tsx docs](https://tsx.hirok.io/watch-mode) (and related search results) say that `watch` should automatically detect changes in _any_ files imported by the entry point (`src/app.ts` in our case). But it clearly is not doing that.

It works if we brute-force it to watch all `.ts` files using the `--include` argument, so I'll just do that.

test-server/src/operation.ts
cmoesel: Ah. Even though I called `dotenv.config()` first in `app.ts`, all of the imports actually load first -- and the line that reads the env in `operation.ts` gets executed when `operation.ts` is imported.

I didn't notice this because I was testing mostly via `USE_TRANSLATION_SERVICE=true npm run dev`.

I've fixed it locally. It turns out you can `import 'dotenv/config'` to load the `.env` config immediately during import.

cmoesel:
Thanks for the excellent review comments, @elsaperelli. I've accepted the typo suggestions in b5f0daa and implemented fixes for the others in f92c40d. Thanks for keeping me honest. I guess I was rushing to get things done before my vacation more than I realized!

cmoesel:
@hossenlopp - Good point regarding `cql-exec-fhir`. I thought that `cql-execution` was declared as a peer dependency on that so it would just use whichever one is already in the dependencies, but I never confirmed via the `package-lock.json`. I'll take a look at that.

IIRC, I did try to do a "normal" import via `../` first but I just could not get it to work with the TypeScript compiler. Perhaps I will give it another go, because I really would prefer that it work that way anyway.

cmoesel:
@hossenlopp - Thanks for noticing the issue with `cql-exec-fhir`. TBH, I misunderstood your comment at first (see above), but I understand now. As it turns out, the test-server doesn't need to use `cql-exec-fhir` at all because the tests are data-model-agnostic. 6c30420 removes the `cql-exec-fhir` dependency and uses the simpler `PatientSource` from `cql-execution` instead. This resulted in a very happy result: _way-improved performance!_ The test runner used to take about 1m 25s to run. Now it takes about 10s to run! Woah!

I also investigated what it would take to have the test-server rely on the cql-execution source .ts files directly (instead of relying on compiled source in `lib`). This was quite tricky, but Codex eventually figured it out. By importing the ts files directly, the test-server can watch the cql-execution source for changes and hot-reload them. That's pretty cool, but it comes at a cost: the configuration is a bit more complex. I'm not sure if hot-loading cql-execution from the test-server is worth the additional complexity or not. For now I put it in a separate branch. Please take a look at [this commit](https://github.com/cqframework/cql-execution/commit/154abb7094266ac4fea821bcbfbbd1a277766def) and let me know if you think we should adopt it or leave things as-is (you too, @elsaperelli). If we leave things as-is, it means you need to manually stop and restart the test server if you make any changes in `cql-execution` source.

hossenlopp:
@cmoesel Replying to your work on getting the hot-reload setup. It looks like that approach could be slightly fragile. I don't make use of the watch re-run features. I feel like if I was iterating on a fix, it makes more sense to be testing on the tests in the core package instead of the cql-test-server based tests.

test-server/src/convert/convert.ts
```diff
@@ -0,0 +1,387 @@
...
+      return toBooleanParameter(name, result);
+    case 'System.Integer':
+      return toIntegerParameter(name, result);
+    case 'System.Long':
+      return toLongParameter(name, result);
+    case 'System.Decimal':
+      return toDecimalParameter(name, result);
+    case 'System.Date':
+      return toDateParameter(name, result);
+    case 'System.DateTime':
```
elsaperelli, line 116: Do we need a return here? I am not sure if it falls under the ranges case though, so maybe this comment can be ignored :)

elsaperelli (approved):
Just one comment, but I am unsure of it so I will go ahead and approve! Woohoo!

test-server/src/convert/convert.ts
cmoesel: This is as intended. CQL Execution uses the `DateTime` class for both CQL DateTime and Time -- so both the `System.DateTime` and `System.Time` cases use the same conversion function.

elsaperelli (approved):

cmoesel:
I've addressed @hossenlopp's comments, but since he is currently unavailable to re-review, I will merge with @elsaperelli's recent approval. Thanks @hossenlopp and @elsaperelli!

cmoesel:
Oh wait! I forgot to add GitHub actions! I will do that first.

PR #361, opened 2026-04-23
Implement CQL Long

cmoesel:
This PR introduces support for CQL Long, represented using `BigInt` in JavaScript. Since Integer and Decimal are represented using Number, we can easily distinguish CQL Longs from Integers and Decimals by their type (`typeof var === 'bigint'`). This actually made implementation easier (and more correct) compared to my initial implementation using Number (which I thought would be easier since it was more of an incremental change, but maybe not).

I based this PR on the `test-server` branch so I could test it with the CQL Test Runner, but in hindsight, that also wasn't that smart because of [cql-tests-runner#79](https://github.com/cqframework/cql-tests-runner/issues/79) (which basically reports that the test runner doesn't correctly detect equal Longs). Oh well. If you want to test it using the Test Runner anyway, feel free -- but I've updated the cql-tests-runner [mitre-wip](https://github.com/cqframework/cql-tests-runner/tree/mitre-wip) branch to skip the Long tests since the runner doesn't support them correctly.

This PR is configured to go into the `test-server` branch, but that's just so it is easy to see only the relevant changes in the diff. I'm marking this draft for now because I think `test-server` should be merged first (_if_ we decide to merge it) and then this should be changed to be based on `main`.

**Submitter:**

**Reviewer:**

Name:


cmoesel:
Closing this PR in favor of #363, which is based on `main` instead of the `test-server` branch.

PR #366, opened 2026-05-29
Fix overlaps for boundless and unknown intervals

### LINKED ISSUE #364: Incorrect Test Patient Evaluation with Interval[null, null] Overlap
As documented in https://oncprojectracking.healthit.gov/support/browse/CQLIT-572, Users are expecting to have a "Cancer Related Pain" diagnosis with a prevalence period that has a null start and end evaluate to TRUE

Image

Verified MADiE application and the interval is passed as a closed interval with low and high as null fields. The expectation is that Closed Interval should be considered as `Minimum start and end, effectively, "for all time"`. So in this example the clause of `where CancerPain.prevalencePeriod overlaps day of Encounter.relevantPeriod` should be evaluated to TRUE ( since it overlays the Encounter Inpatient relavant period ).
```
Prevalence Period from test Patient
{
    "low": null,
    "high": null,
    "lowClosed": true,
    "highClosed": true
}
```

Additional Notes:
https://github.com/cqframework/CQL-Formatting-and-Usage-Wiki/blob/master/Source/Cooking%20With%20CQL/101/01-IntervalNullNull.md

Attached cqmMeasure and Synthetic Patient below

[cqmMeasure.txt](https://github.com/user-attachments/files/28362582/cqmMeasure.txt)

[testPatient.json](https://github.com/user-attachments/files/28362584/testPatient.json)

Options used 
```
{
    "doPretty": true,
    "includeClauseResults": true
}
```

cmoesel:
Add special case logic for `overlaps`, `overlaps before`, and `overlaps after` when arguments are boundless intervals (e.g., `Interval[null, null]`) or  unknown intervals (e.g., `Interval(null, null)`).

When we process an Interval that has both boundaries null, we can't infer the point type since we have no value to infer it from. This caused problems in the existing logic that needs to know the point type for some calculations. To address this, I've added special-case logic for these edge cases since we often don't really need to know the point type to arrive at the correct answer.

My [CQL Long PR](https://github.com/cqframework/cql-execution/pull/363) introduces the approach of using the `resultType` from the ELM when we can't easily infer types from the data. We should extend that to also work in these cases, but since we want to get this fix out ASAP, I decided to implement this approach first. We need this approach regardless since ELM does not always carry the `resultType`.

_NOTE: The CI that runs `npm audit` currently fails, but the reported audit issues are currently not resolvable, so there's not much we can do._

Fixes #364.

**Submitter:**

**Reviewer:**

Name:


elsaperelli (approved):
🥳

PR #367, opened 2026-06-02
Safe import for immutable

elsaperelli:

## Summary
This PR updates the `immutable` imports that were causing issues down the line when cql-execution is used as a dependency in a browser-bundling application. This was not clear with Node through CommonJS because it worked, but the Immutable ESM build that bundled browser environments use exposes `Set` (and others) as named exports. Therefore, this PR fixes errors that `Set` was undefined.

**Submitter:**

**Reviewer:**

Name:

### DIFF
--- a/examples/browser/cql4browsers.js
+++ b/examples/browser/cql4browsers.js
@@ -5715,9 +5715,6 @@ exports.Library = Library;
 
 },{"./expressions":23}],28:[function(require,module,exports){
 "use strict";
-var __importDefault = (this && this.__importDefault) || function (mod) {
-    return (mod && mod.__esModule) ? mod : { "default": mod };
-};
 Object.defineProperty(exports, "__esModule", { value: true });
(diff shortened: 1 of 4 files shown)


cmoesel:
Thanks for taking care of this, @elsaperelli. Tricky!

I just pushed [2043b29](https://github.com/cqframework/cql-execution/pull/367/commits/2043b2993088013c305d6dd5e09af3190b08f9ab) with a couple more tweaks:
* Fixed use of `Immutable.Collection` in `immutableUtil.ts`. Apparently this didn't cause issues in the runtime code but could cause issues in TypeScript compilation for downstream projects who import `immutableUtil.ts` (which is, admittedly, not likely, but... how was `Immutable.Collection` even working at all after you got rid of the root `Immutable` import?).
* Switched to the new import syntax in `immutableUtil-test.ts`. This probably was not necessary (since it's not a production file and it's always run in Node), but since I was already making a change in this PR, I thought this might be good too so we have a consistent approach throughout the whole project.

cmoesel (approved):
Looks good. Thank you, @elsaperelli

Issue #291, opened 2022-10-27
Support ProperContains/ProperIn for Lists


cmoesel:
The [ProperContains](https://cql.hl7.org/04-logicalspecification.html#proper-contains) / [ProperIn](https://cql.hl7.org/04-logicalspecification.html#proper-in) operators are similar to [ProperIncludes](https://cql.hl7.org/04-logicalspecification.html#proper-includes) and [ProperIncludedIn](https://cql.hl7.org/04-logicalspecification.html#proper-included-in). The primary difference is that the former two support relationships between a list and a point, while the latter two support relationships between two lists.

Note that despite their names, these operates are invoked in CQL using the same syntax is `ProperIncludes`/`ProperIncludedIn`.  E.g.
```
define ListProperlyContainsPoint: {1, 2, 3, 4, 5} properly includes 3
define PointProperlyInList: 3 properly included in {1, 2, 3, 4, 5}
```

dehall:
Implemented in #369

PR #369, opened 2026-06-30
Implement ProperContains/ProperIn for both List and Interval

dehall:
This PR implements the [ProperContains](https://cql.hl7.org/04-logicalspecification.html#proper-contains) and [ProperIn](https://cql.hl7.org/04-logicalspecification.html#proper-in) operators, for Lists (implements #291) and Intervals (implements #292)
As of today, the spec reads as follows:
> The ProperContains operator returns true if the first operand properly contains the second.
> 
> There are two overloads of this operator: List, T: The type of T must be the same as the element type of the list. Interval, T : The type of T must be the same as the point type of the interval.
> 
> For the List, T overload, this operator returns true if the given element is in the list, and it is not the only element in the list, using equality semantics, with the exception that null elements are considered equal. If the first argument is null, the result is false. If the second argument is null, the result is true if the list contains any null elements and at least one other element, and false otherwise.
> 
> For the Interval, T overload, this operator returns true if the given point is greater than the starting point of the interval, and less than the ending point of the interval, as determined by the Start and End operators. If precision is specified and the point type is a Date, DateTime, or Time type, comparisons used in the operation are performed at the specified precision. If the first argument is null, the result is false. If the second argument is null, the result is null.

(ProperIn is just the inverse - ProperContains is `myList properly includes myItem`, ProperIn is `myItem properly included in myList`)

However, there is an upcoming clarification to the spec that will clarify two aspects of these operators for Lists:
1. that the list membership operators will [always] return true or false (i.e. we add "otherwise the result is false" to clarify)
2. that the proper list operators (membership and otherwise) have list, not set, semantics.

This PR is implemented per these clarifications.

Note also that the cql tests are being updated as part of that clarification, see https://github.com/cqframework/cql-tests/pull/119 , so I have copied the latest from that PR in here. As of now that PR has a couple approvals already but if it changes again I'll re-update. 

Fortunately, the spec is pretty unambiguous for the Interval operators.

**Submitter:**

**Reviewer:**

Name:

### DIFF
--- a/src/datatypes/interval.ts
+++ b/src/datatypes/interval.ts
@@ -114,6 +114,28 @@ export class Interval {
     );
   }
 
+  properContains(item: any, precision?: any) {
+    if (item != null && item.isInterval) {
+      throw new Error('Argument to contains must be a point');
+    }
+    let lowFn;
+    if (this.lowClosed && this.low == null) {
+      lowFn = () => true;
+    } else {
+      lowFn = cmp.lessThan;
+    }
+    let highFn;
+    if (this.highClosed && this.high == null) {
+      highFn = () => true;
+    } else {
+      highFn = cmp.greaterThan;
+    }
+    return ThreeValuedLogic.and(
+      lowFn(this.low, item, precision),
+      highFn(this.high, item, precision)
+    );
+  }
+
   properlyIncludes(other: any, precision?: any) {
     if (other == null || !other.isInterval) {
       throw new Error('Argument to properlyIncludes must be an interval');
--- a/src/datatypes/interval.ts
+++ b/src/datatypes/interval.ts
@@ -115,7 +115,9 @@ export class Interval {
   }
 
   properContains(item: any, precision?: any) {
-    if (item != null && item.isInterval) {
+    if (item == null) {
+      return null;
+    } else if (item.isInterval) {
       throw new Error('Argument to contains must be a point');
     }
     return ThreeValuedLogic.and(
--- a/src/datatypes/interval.ts
+++ b/src/datatypes/interval.ts
@@ -118,21 +118,9 @@ export class Interval {
     if (item != null && item.isInterval) {
       throw new Error('Argument to contains must be a point');
     }
-    let lowFn;
-    if (this.lowClosed && this.low == null) {
-      lowFn = () => true;
-    } else {
-      lowFn = cmp.lessThan;
-    }
-    let highFn;
-    if (this.highClosed && this.high == null) {
-      highFn = () => true;
-    } else {
-      highFn = cmp.greaterThan;
-    }
     return ThreeValuedLogic.and(
-      lowFn(this.low, item, precision),
-      highFn(this.high, item, precision)
+      cmp.lessThan(this.start(), item, precision),
+      cmp.greaterThan(this.end(), item, precision)
     );
   }
 
--- a/src/elm/overloaded.ts
+++ b/src/elm/overloaded.ts
@@ -269,6 +269,54 @@ export class ProperIncludedIn extends Expression {
   }
 }
 
+export class ProperIn extends Expression {
+  precision?: any;
+
+  constructor(json: any) {
+    super(json);
+    this.precision = json.precision != null ? json.precision.toLowerCase() : undefined;
+  }
+
+  async exec(ctx: Context) {
+    const [item, container] = await this.execArgs(ctx);
+    if (container == null) {
+      return false;
+    }
+    if (typeIsArray(container)) {
+      return LIST.doProperContains(container, item);
+    } else {
+      if (item == null) {
+        return null;
+      }
+      return IVL.doProperContains(container, item, this.precision);
+    }
+  }
+}
+
+export class ProperContains extends Expression {
+  precision?: any;
+
+  constructor(json: any) {
+    super(json);
+    this.precision = json.precision != null ? json.precision.toLowerCase() : undefined;
+  }
+
+  async exec(ctx: Context) {
+    const [container, item] = await this.execArgs(ctx);
+    if (container == null) {
+      return false;
+    }
+    if (typeIsArray(container)) {
+      return LIST.doProperContains(container, item);
+    } else {
+      if (item == null) {
+        return null;
+      }
+      return IVL.doProperContains(container, item, this.precision);
+    }
+  }
+}
+
 export class Length extends Expression {
   constructor(json: any) {
     super(json);
--- a/src/elm/overloaded.ts
+++ b/src/elm/overloaded.ts
@@ -285,9 +285,6 @@ export class ProperIn extends Expression {
     if (typeIsArray(container)) {
       return LIST.doProperContains(container, item);
     } else {
-      if (item == null) {
-        return null;
-      }
       return IVL.doProperContains(container, item, this.precision);
     }
   }
@@ -309,9 +306,6 @@ export class ProperContains extends Expression {
     if (typeIsArray(container)) {
       return LIST.doProperContains(container, item);
     } else {
-      if (item == null) {
-        return null;
-      }
       return IVL.doProperContains(container, item, this.precision);
     }
   }
--- a/test/datatypes/interval-test.ts
+++ b/test/datatypes/interval-test.ts
@@ -2,7 +2,12 @@ import should from 'should';
 import { DateTime, MIN_DATETIME_VALUE, MAX_DATETIME_VALUE } from '../../src/datatypes/datetime';
 import { Interval } from '../../src/datatypes/interval';
 import { Uncertainty } from '../../src/datatypes/uncertainty';
-import { ELM_DECIMAL_TYPE } from '../../src/util/elmTypes';
+import {
+  ELM_DATETIME_TYPE,
+  ELM_DECIMAL_TYPE,
+  ELM_INTEGER_TYPE,
+  ELM_LONG_TYPE
+} from '../../src/util/elmTypes';
 import data from './interval-data';
 
 const xy = (obj: any) => [obj.x, obj.y];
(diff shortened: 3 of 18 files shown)


src/datatypes/interval.ts
```diff
@@ -105,6 +105,16 @@ export class Interval {
     );
   }
 
+  properContains(item: any, precision?: any) {
+    if (item != null && item.isInterval) {
+      throw new Error('Argument to contains must be a point');
+    }
```
cmoesel: The [spec says](https://build.fhir.org/ig/HL7/cql/04-logicalspecification.html#proper-contains):
> If the second argument is null, the result is null.

_(note: in this context, `item` is the second argument)._

Although a `null` item should cause the three-valued logic on lines 112-115 to return `null` anyway, it might be more straight-forward and obviously aligned w/ the spec if we made it explicit. We need to do a null-check anyway, so consider this:
```suggestion
    if (item == null) {
      return null;
    } else if (item.isInterval) {
      throw new Error('Argument to contains must be a point');
    }
```

cmoesel: I see now that you do the explicit null-check in `src/elm/overloaded.ts` lines 287-289 and 311-313, but since `Interval.properContains` should also stand on its own, I think it's still worth considering the above.

src/elm/overloaded.ts
```diff
@@ -268,6 +268,54 @@ export class ProperIncludedIn extends Expression {
...
+      return false;
+    }
+    if (typeIsArray(container)) {
+      return LIST.doProperContains(container, item);
+    } else {
+      if (item == null) {
+        return null;
+      }
+      return IVL.doProperContains(container, item, this.precision);
+    }
```
cmoesel, line 289: It seems we're inconsistent in this file regarding our approach to these types of overloads. While `contains` and `in` take the approach you've mimicked above, `Includes`, `IncludedIn`, `ProperIncludes`, and `ProperIncludedIn` don't do a null-check on `item` here since it's handled in the respective interval/list implementations. This allows them to do something like this instead:
```suggestion
    const lib = typeIsArray(container) ? LIST : IVL;
    return lib.doProperContains(container, item, this.precision);
```

That said, I'm a little conflicted on that approach too since it takes advantage of JS's loose approach to function signatures (given that `LIST.doProperContains` technically does not take a 3rd argument).

Anyway, I guess I'm saying this doesn't have to change, but I'm interested in your thoughts on it.

src/elm/overloaded.ts
```diff
@@ -268,6 +268,54 @@ export class ProperIncludedIn extends Expression {
...
```
cmoesel, line 310: See comment on lines 284 - 291 above.

test/datatypes/interval-test.ts
```diff
@@ -110,6 +110,86 @@ describe('DateTimeInterval.contains', () => {
...
+    d = data();
+  });
+
+  it('should properly calculate dates before it', () => {
+    d.all2012.closed.properContains(d.bef2012.full).should.be.false();
+  });
+
+  it('should properly calculate the left boundary date', () => {
+    d.all2012.closed.properContains(d.beg2012.full).should.be.false();
+    d.all2012.open.properContains(d.beg2012.full).should.be.false();
```
cmoesel: Should we add a just barely true case too? E.g.
```ts
d.all2012.closed.properContains(d.beg2012.open).should.be.true();
```

test/datatypes/interval-test.ts
```diff
@@ -110,6 +110,86 @@ describe('DateTimeInterval.contains', () => {
...
+  });
+
+  it('should properly calculate dates in the middle of it', () => {
+    d.all2012.closed.properContains(d.mid2012.full).should.be.true();
+  });
+
+  it('should properly calculate the right boundary date', () => {
+    d.all2012.closed.properContains(d.end2012.full).should.be.false();
+    d.all2012.open.properContains(d.end2012.full).should.be.false();
+  });
```
cmoesel: Should we add a just barely true case too? E.g.,
```ts
d.all2012.closed.properContains(d.end2012.open).should.be.true();
```

test/datatypes/interval-test.ts
```diff
@@ -110,6 +110,86 @@ describe('DateTimeInterval.contains', () => {
...
+    new Interval(null, date).properContains(late).should.be.false();
+    new Interval(null, date, false, true).properContains(date).should.be.false();
+    should(new Interval(null, date, false, true).properContains(early)).be.null();
+    new Interval(null, date, false, true).properContains(late).should.be.false();
+    new Interval(date, null).properContains(late).should.be.true();
+    new Interval(date, null).properContains(early).should.be.false();
+    new Interval(date, null).properContains(maxDate).should.be.false();
+    new Interval(date, null, true, false).properContains(date).should.be.false();
+    should(new Interval(date, null, true, false).properContains(late)).be.null();
+    new Interval(date, null, true, false).properContains(early).should.be.false();
```
cmoesel: Should we also add tests for the unbounded and unknown intervals? E.g.:
```ts
new Interval(null, null, true, true, ELM_DATETIME_TYPE).properContains(date).should.be.true();
should(new Interval(null, null, false, false, ELM_DATETIME_TYPE).properContains(date)).be.null();
```
I'm not sure if they will work or not since we have a separate unmerged PR already for handling unbounded and unknown intervals. If they don't work and you want to defer until that PR, that's OK.

test/datatypes/interval-test.ts
```diff
@@ -110,6 +110,86 @@ describe('DateTimeInterval.contains', () => {
...
+
+  it('should properly handle imprecision', () => {
+    d.all2012.closed.properContains(d.bef2012.toMonth).should.be.false();
+    should.not.exist(d.all2012.closed.properContains(d.beg2012.toMonth));
+    d.all2012.closed.properContains(d.mid2012.toMonth).should.be.true();
+    should.not.exist(d.all2012.closed.properContains(d.end2012.toMonth));
+    d.all2012.closed.properContains(d.aft2012.toMonth).should.be.false();
+
+    d.all2012.toMonth.properContains(d.bef2012.toMonth).should.be.false();
+    d.all2012.toMonth.properContains(d.beg2012.toMonth).should.be.false();
```
cmoesel: I feel like this result is a little surprising, but it is correct according to spec since Before/After say:

> If no precision is specified, the comparison is performed beginning with years (or hours for time values) and proceeding to the finest precision specified in either input.

In this case, since both are defined to months, precision stops there leaving us with `2012-01 < 2012-01` (false). So... OK.

test/datatypes/interval-test.ts
```diff
@@ -1890,6 +1970,83 @@ describe('IntegerInterval.contains', () => {
...
+    d = data();
+  });
+
+  it('should properly calculate integers less than it', () => {
+    d.zeroToHundred.closed.properContains(-5).should.be.false();
+  });
+
+  it('should properly calculate the left boundary integer', () => {
+    d.zeroToHundred.closed.properContains(0).should.be.false();
+    d.zeroToHundred.open.properContains(0).should.be.false();
```
cmoesel: Should we add a just barely `true` test case here too? And maybe a just another variation of the `false` as well? E.g.,
```ts
d.zeroToHundred.closed.properContains(1).should.be.true();
d.zeroToHundred.open.properContains(1).should.be.false();
```

test/datatypes/interval-test.ts
```diff
@@ -1890,6 +1970,83 @@ describe('IntegerInterval.contains', () => {
...
+  });
+
+  it('should properly calculate integers in the middle of it', () => {
+    d.zeroToHundred.closed.properContains(50).should.be.true();
+  });
+
+  it('should properly calculate the right boundary integer', () => {
+    d.zeroToHundred.closed.properContains(100).should.be.false();
+    d.zeroToHundred.open.properContains(100).should.be.false();
```
cmoesel: Should we add a just barely true test case here too? And maybe a just another variation of the false as well? E.g.,
```ts
d.zeroToHundred.closed.properContains(99).should.be.true();    
d.zeroToHundred.open.properContains(99).should.be.false();
```

test/datatypes/interval-test.ts
```diff
@@ -1890,6 +1970,83 @@ describe('IntegerInterval.contains', () => {
...
+    new Interval(null, 0).properContains(-123456789).should.be.true();
+    new Interval(null, 0).properContains(1).should.be.false();
+    new Interval(null, 0, false, true).properContains(0).should.be.false();
+    should(new Interval(null, 0, false, true).properContains(-123456789)).be.null();
+    new Interval(null, 0, false, true).properContains(1).should.be.false();
+    new Interval(0, null).properContains(123456789).should.be.true();
+    new Interval(0, null).properContains(-1).should.be.false();
+    new Interval(0, null, true, false).properContains(0).should.be.false();
+    should(new Interval(0, null, true, false).properContains(123456789)).be.null();
+    new Interval(0, null, true, false).properContains(-1).should.be.false();
```
cmoesel: Should we also add tests for the unbounded and unknown intervals? E.g.:
```ts
new Interval(null, null, true, true, ELM_INTEGER).properContains(0).should.be.true();
should(new Interval(null, null, false, false, ELM_INTEGER).properContains(0)).be.null();
```
I'm not sure if they will work or not since we have a separate unmerged PR already for handling unbounded and unknown intervals. If they don't work and you want to defer until that PR, that's OK.

cmoesel (commented):
This looks good, @dehall. The implementation is straight-forward and the tests are robust. I've left a few small comments for your consideration. This also needs to be rebased (and it looks like there are conflicts).

src/elm/overloaded.ts
dehall: It turns out this suggestion doesn't compile as-is. That's obviously not insurmountable, but thinking about why it doesn't work makes me lean (slightly) toward explicitly calling the appropriate class vs using the "implied interface" version.

The error is:

```
src/elm/overloaded.ts:286:50 - error TS2554: Expected 2 arguments, but got 3.

286     return lib.doProperContains(container, item, this.precision);
                                                     ~~~~~~~~~~~~~~
```

Turns out this is because of how the 2 functions are defined and how TypeScript resolves the type of `lib`:

```typescript
  // list.ts
  doProperContains(container: any[], item: any)

  // interval.ts
  doProperContains(interval: any, item: any, precision?: any)
```

If we change `list.doProperContains` :: `container: any[]` to `container: any` then it will work. But I don't love the idea of widening the type of a parameter just for that. And if at some point we want to continue making function signatures more explicitly-typed (eg, interval `doProperContains` does not really accept just anything for the first parameter, it requires an `Interval`) then I think the if/else version will be easier to manage.

All that to say, I don't have super strong feelings. In general I like less magic. But I'm also happy to change this or any of the other instances in this file for consistency.

test/datatypes/interval-test.ts
dehall: The first one currently returns null so I'd lean toward deferring here. (The second one also returns null but I don't think it's for the right reasons.) I can add these with a skip for now

src/elm/overloaded.ts
cmoesel: Let's not widen the type in the signature. If the magic worked as-is, I'd be cool with the shortcut since it is consistent w/ other operator implementations. But I don't want to move even further away from strict typing just to make that work.

src/datatypes/interval.ts
dehall: Updated in https://github.com/cqframework/cql-execution/pull/369/commits/ac816342ecd24074bdfe56019ced64486edbcd2a

test/datatypes/interval-test.ts
dehall: Added in https://github.com/cqframework/cql-execution/pull/369/commits/84c462c4d4b256507c7b376721af949d5fe25e2e
Note that I used the successor function here which I half feel is cheating but for dates it's not as clean as integers/longs

test/datatypes/interval-test.ts
dehall: Also added in https://github.com/cqframework/cql-execution/pull/369/commits/84c462c4d4b256507c7b376721af949d5fe25e2e

test/datatypes/interval-test.ts
dehall: Also added in https://github.com/cqframework/cql-execution/pull/369/commits/84c462c4d4b256507c7b376721af949d5fe25e2e

test/datatypes/interval-test.ts
dehall: Also added in https://github.com/cqframework/cql-execution/pull/369/commits/84c462c4d4b256507c7b376721af949d5fe25e2e

test/datatypes/interval-test.ts
dehall: Added (skipped) in https://github.com/cqframework/cql-execution/pull/369/commits/44a25f0014fa889d6d445418e30edd6d11ada88f

test/datatypes/interval-test.ts
dehall: Added (skipped) in https://github.com/cqframework/cql-execution/pull/369/commits/44a25f0014fa889d6d445418e30edd6d11ada88f

dehall:
Thanks @cmoesel , I think I've addressed all your comments, and rebased plus added LongInterval.properContains tests (https://github.com/cqframework/cql-execution/pull/369/commits/cb5eb5cf27105f4eb60d5937a0e08ff2a486e603)

test/datatypes/interval-test.ts
cmoesel: Ah, right. For some reason I was thinking `d.beg2012` referenced intervals, but it references dates, so `.open` doesn't work. Using `successor` is _kind of_ cheating, but I think it's fair.

cmoesel (approved):
Thanks, @dehall. This looks good and works well!

src/elm/overloaded.ts
```diff
@@ -269,6 +269,54 @@ export class ProperIncludedIn extends Expression {
...
+
+  async exec(ctx: Context) {
+    const [item, container] = await this.execArgs(ctx);
+    if (container == null) {
+      return false;
+    }
+    if (typeIsArray(container)) {
+      return LIST.doProperContains(container, item);
+    } else {
+      if (item == null) {
```
lmd59: Is this check redundant with the check in src/datatypes/interval.ts line [118](https://github.com/cqframework/cql-execution/pull/369/changes#diff-9381d2bf7dbbdb8f1cefadb5b716744cee39d3c5dc27dafb61808a8e7e4b347aR118)?

src/elm/overloaded.ts
```diff
@@ -269,6 +269,54 @@ export class ProperIncludedIn extends Expression {
...
+
+  async exec(ctx: Context) {
+    const [container, item] = await this.execArgs(ctx);
+    if (container == null) {
+      return false;
+    }
+    if (typeIsArray(container)) {
+      return LIST.doProperContains(container, item);
+    } else {
+      if (item == null) {
```
lmd59: Same as above (is this check redundant with check in datatypes/interval.ts?)

lmd59 (commented):
After the discussion to get here, I appreciate the simplicity of the final solution. Looking good. I had one question (two related comments).

src/elm/overloaded.ts
dehall: Yes it is, removed this and the other one in https://github.com/cqframework/cql-execution/pull/369/commits/13b7fb50c140fe410d1082bfde571d21a44a7366

lmd59 (approved):

cmoesel (approved):
Still looks good to me!

PR #373, opened 2026-07-21
Interval Operator Overhaul

cmoesel:
This PR replaces #365 by aligning (almost) all interval operators more closely with the exact language and algorithms prescribed in the spec and using the `Start` and `End` operators to determine interval boundaries. In addition, interval point types are detected more accurately (using `resultType` where possible) to ensure accurate interval representations when both boundaries are null. As a result, handling of closed null and open null boundaries has been greatly improved.

Other tangentially related improvements have been made throughout to address additional issues that arose in implementation and testing of the interval operators, especially in regard to arithmetic functions, precision, and data type limits (min/max).

Note:
* Interval expand and collapse are quite complex. Refactoring these operators should be addressed in a future PR.
* The spec has diverging definitions of how properly contains should work. I've refactored it to work how I _think_ it should work, but I've done it in a separate commit in case we need to back that out. Check the [jira issue](https://jira.hl7.org/browse/FHIR-57890) and [zulip conversation](https://chat.fhir.org/#narrow/channel/179220-cql/topic/Discrepancy.20in.20ProperContains.20.2F.20Properly.20Includes/near/611720760) for updates.  

**Submitter:**

**Reviewer:**

Name:


cmoesel:
@lmd59 - I added you as a reviewer since you reviewed #365 (which this replaces), but if you don't have time, then I think anyone else on the team would be fine. @dehall will also review, but this one is kind of big, so two reviewers might be good -- especially if someone from your team can regress it with fqm.

src/datatypes/datetime.ts
```diff
@@ -550,7 +558,7 @@ abstract class AbstractDate {
       case 'millisecond':
         return 999;
       default:
-        throw new Error('Tried to clieling a field that has no cieling value: ' + field);
+        throw new Error('Tried to ceiling a field that has no cieling value: ' + field);
```
dehall: Are you open to fixing the name of this whole function while you're here?

test/spec-tests/skip-list.txt
```diff
@@ -62,5 +58,4 @@ CqlArithmeticFunctionsTest.Modulo.Modulo10By3Quantity                 Modulo not
 "CqlArithmeticFunctionsTest.Truncated Divide.TruncatedDivide10By5DQuantity"           Truncated divide not implemented for Quantity
 "CqlArithmeticFunctionsTest.Truncated Divide.TruncatedDivide414By206DQuantity"        Truncated divide not implemented for Quantity
 
-# Unimplemented (New in CQL 2.0)
-CqlListOperatorsTest.Slice                              Slice not implemented
```
dehall, line 66: This looks like an incorrect change by the `check-skip-list` script, similar to what happened to me on a recent PR. The cql-to-elm-cli failed to generate ELM for `CqlListOperatorsTest.cql` but doesn't report the error by a nonzero exit code, and so the `check-skip-list` task happily continues, runs the tests against the old ELM, and thinks the new tests are 100% passing, so a test that really should be skipped gets un-skipped. 

The other way to notice this is that this PR currently has a modified `test/spec-tests/cql/CqlListOperatorsTest.cql` but not a corresponding `CqlListOperatorsTest.json`. Shoutout to @lmd59 for catching that on my earlier PR so now I know what to look for. I have a PR up on the translator CLI to help here, hopefully that gets some eyes on it soon.

Turns out, this specific Slice test case is what's causing the translation to fail:

```
================================================================================
TRANSLATE /Users/dehall/cql-execution-review/test/generator/../spec-tests/cql/CqlListOperatorsTest.cql
Translation failed due to errors:
[... omitted a bunch of warnings]
Error:[946:17, 946:40] Could not resolve call to operator Slice with signature (list<System.Integer>).
```

I think for now we just want to put this skip line back, the rest of the changes in this file should be ok.

test/datatypes/date-test.ts
```diff
@@ -196,6 +196,36 @@ describe('Date.add', () => {
   });
 });
 
+describe('Date.successor', () => {
+  it('should use the specified precision', () => {
+    const date = Date.parse('2000-06-15');
+    date.successor(Date.Unit.YEAR).should.eql(Date.parse('2001-06-15'));
+    date.successor(Date.Unit.MONTH).should.eql(Date.parse('2000-07-15'));
+    date.successor(Date.Unit.DAY).should.eql(Date.parse('2000-06-16'));
```
dehall, line 204: Just an observation, not a request for change at this point - I'm surprised there aren't more leap year tests in this file. I realize date logic is handled by the underlying luxon library, but still

src/util/math.ts
```diff
@@ -279,9 +352,10 @@ export function maxValueForInstance(val: any) {
   } else if (val && val.isDate) {
     return MAX_DATE_VALUE?.copy();
   } else if (val && val.isQuantity) {
-    const val2 = val.clone();
-    val2.value = maxValueForInstance(val2.value);
-    return val2;
+    // Although the spec says max Quantity has unit '1', it doesn't make sense to change the unit,
```
dehall: The comment & approach here make sense but if we're intentionally going against what the spec says, is there a zulip discussion or jira ticket about this as well?

dehall (commented):
I haven't had time to fully finish reviewing but overall it looks good. Submitting the few comments I do have now, since I'll be out Monday

test/datatypes/date-test.ts
cmoesel: Yes, there are a few leap year tests, but there certainly could be more. Since most of this would happen in the date/datetime tests, however, it's probably outside the scope of this (mostly) interval-focused PR.

src/util/math.ts
cmoesel: OK, yeah, that's a fair point. There is actually already some language about a similar approach in [predecessor](https://build.fhir.org/ig/HL7/cql/09-b-cqlreference.html#predecessor) and [successor](https://build.fhir.org/ig/HL7/cql/09-b-cqlreference.html#successor), but I think the `Interval`, `Start`, and `End` specifications need to also provide a special-case for Quantity intervals. See: https://jira.hl7.org/browse/FHIR-57935

src/datatypes/datetime.ts
cmoesel: Even better... it turns out that function is no longer used because I majorly refactored the functions that used to call it. So I just removed it and some similar unused functions in [7d2f99d](https://github.com/cqframework/cql-execution/pull/373/commits/7d2f99d5953ef7faab677e16163297a3e3e5498a).

test/spec-tests/skip-list.txt
cmoesel: You're right. I put it back in [d8d8cdc](https://github.com/cqframework/cql-execution/pull/373/commits/d8d8cdca12083f632783f237c3e129d07151ae9f).

src/util/math.ts
cmoesel: I added the Jira link to the comments in [11111d5](https://github.com/cqframework/cql-execution/pull/373/commits/11111d5f1c6da9a8fff0645ee3ddd7ce8cd11626), which also includes other related changes.

cmoesel:
@dehall and @lmd59 - this is ready for re-review. I also updated all of the dependencies that could be updated and documented the others in `DEPENDENCY-NOTES.md`.

src/datatypes/interval.ts
```diff
@@ -78,635 +73,864 @@ export class Interval {
...
-      (other.high == null && !other.highClosed)
+      this.pointType === ELM_DATE_TYPE ||
+      this.pointType === ELM_DATETIME_TYPE ||
+      this.pointType === ELM_TIME_TYPE
     ) {
-      return null;
+      const boundaries = [this.low, this.high, other.low, other.high];
+      const leastPreciseBoundary = boundaries.reduce((least, boundary) =>
+        boundary?.isLessPrecise(least) ? boundary : least
+      );
```
dehall: Looks like this can throw if `this.low` is null:

```
const ia = new Interval(null, DateTime.parse('2012-12-31'));
const ib = new Interval(DateTime.parse('2012-06-01'), DateTime.parse('2013-01-01'));
ia.except(ib)
```
--> 
```
TypeError: Cannot read properties of null (reading 'year')
      at DateTime.isSamePrecision (src/datatypes/datetime.ts:149:39)
      at DateTime.isLessPrecise (src/datatypes/datetime.ts:136:18)
      at /Users/dehall/cql-execution-review/src/datatypes/interval.ts:313:19
```

or when it's an uncertainty:

```
 const a = new Interval(
   new Uncertainty(DateTime.parse('2012-01-01'), DateTime.parse('2012-01-31')),
   DateTime.parse('2012-12-31')
 );

 const b = new Interval(DateTime.parse('2012-06-01'), DateTime.parse('2013-01-01'));

  a.except(b);
```
-->
```
TypeError: leastPreciseBoundary?.getPrecision is not a function
      at Interval.except (src/datatypes/interval.ts:315:41)
```

src/datatypes/interval.ts
```diff
@@ -78,635 +73,864 @@ export class Interval {
       newHigh = this.high.copy();
     }
 
-    return new Interval(newLow, newHigh, this.lowClosed, this.highClosed);
+    return new Interval(newLow, newHigh, this.lowClosed, this.highClosed, this.pointType);
   }
 
+  // https://build.fhir.org/ig/HL7/cql/09-b-cqlreference.html#contains
```
dehall, line 81: I realize it's probably not possible but it would be nice if there was a more permanent version of these URLs we could link to. Linking to the CI build version of the spec just feels wrong

src/datatypes/interval.ts
```diff
@@ -78,635 +73,864 @@ export class Interval {
...
-  return [x, y].every(z => z != null && z.isDateTime);
+// Return an interval that uses the standard null boundary conventions if appropriate:
+// - low closed boundary with min value for the type is changed to null closed boundary
+// - high closed boundary with max value for the type is changed to null closed boundary
+// - low open boundary that is uncertainty from min value of type to max value or interval high
+//   value is changed to null open boundary
+// - high open boundary that is uncertainty from min value of type or interval low value to max
+//   value of type is changed to null open boundary
+//
+// TODO: Is this necessary? Do we care how it is represented if the representations have the same
```
dehall: Is this something we want to think more about before merging this?

src/datatypes/interval.ts
```diff
@@ -78,635 +73,864 @@ export class Interval {
...
-function areNumeric(x: any, y: any) {
-  return [x, y].every(z => {
-    return (
-      typeof z === 'number' ||
-      typeof z === 'bigint' ||
-      (z != null && z.isUncertainty && (typeof z.low === 'number' || typeof z.low === 'bigint'))
+function performConversionIfNecessary(left: Interval, right: Interval) {
+  if (left.pointType === ELM_DATE_TYPE && right.pointType === ELM_DATETIME_TYPE) {
+    left = new Interval(
+      left.low?.getDateTime(),
```
dehall: These calls in this function should support `low` or `high` being an uncertainty as well. Here's a quick snippet to replicate:

```
  const dateInterval = new Interval(
    new Uncertainty(Date.parse('2012-01-01')),
    Date.parse('2012-12-31'),
    true,
    true,
    ELM_DATE_TYPE
  );

  const dateTimeInterval = new Interval(
    new Uncertainty(DateTime.parse('2012-01-01')),
    DateTime.parse('2012-12-31'),
    true,
    true,
    ELM_DATETIME_TYPE
  );

  dateInterval.sameAs(dateTimeInterval);
```

-->

```
TypeError: left.low?.getDateTime is not a function
      at performConversionIfNecessary (src/datatypes/interval.ts:853:17)
      at Interval.sameAs (src/datatypes/interval.ts:346:27)
```

dehall (changes requested):
I found just a couple more places where a call to `.low` or `.high` needs to consider uncertainties to prevent errors, but overall I think this is really close. After those are addressed I'll be good to approve. Two other comments are just on comments, and maybe those are already how we want them/as good as they can be at this point.

src/datatypes/interval.ts
cmoesel: I agree, but the CI build contains all of the updated interval language and is going to be published as CQL 2.0 _very soon_. It seemed even more inappropriate (and against the intent of the PR) to use definitions from the current 1.5.3 CQL publication. I was thinking that once CQL 2.0 is published, we could update all of these using a pretty simple find/replace.

src/datatypes/interval.ts
cmoesel: I kind of wanted to know what others (i.e., the reviewers) think. So... what do you think?

a) Keep it for now.
b) Keep it forever.
c) Get rid of it now.

src/datatypes/interval.ts
cmoesel: Addressed in [af4da33](https://github.com/cqframework/cql-execution/pull/373/commits/af4da33b377b5acd749fc2fd9410cfbf22417ee7).

src/datatypes/interval.ts
cmoesel: I added a TODO in [af4da33](https://github.com/cqframework/cql-execution/pull/373/commits/af4da33b377b5acd749fc2fd9410cfbf22417ee7) to replace the URLs w/ stable URLs once CQL 2.0 is published.

src/datatypes/interval.ts
cmoesel: Addressed in [af4da33](https://github.com/cqframework/cql-execution/pull/373/commits/af4da33b377b5acd749fc2fd9410cfbf22417ee7).

cmoesel:
Thanks for the review, @dehall. Good finds. I've addressed them now, so this is ready for re-review (again!).

src/datatypes/interval.ts
dehall: I guess after looking at this a little closer I'd lean towards remove but I'm totally ok with keeping it for now. It's not strictly necessary, and the fact that it's only used in a few places but not everywhere that returns an interval feels a little inconsistent. In practice the differences shouldn't matter, and it doesn't break anything, but it adds some complexity that isn't required.

dehall (approved):
Looks good! Per the last comment, there's one thing I'd be ok with removing but since it's there and tested I'm also fine with leaving it in.

src/datatypes/interval.ts
cmoesel: My main concern is for others who use our library and inspect (or display) its output. Will they recognize `Interval[-2147483648, 2147483647]` as the same as `Interval[null, null] as Interval<Integer>`? Or `Interval[Uncertainty(-2147483648, 0), 10]` as the same as `Interval(null, 10]`? Normalizing intervals avoids that and keeps it closer to the way our implementation worked before this PR (which is to keep null boundaries intact).

That said, I removed it locally and tested using the [cql-tests-runner](https://github.com/cqframework/cql-tests-runner) and was pleased to see that it did _not_ affect our test results. I'm not sure if it is because the runner is smart or the tests don't exercise any of the operations where this pops up. It did cause a few of our own tests to fail, but that was because my implementation of the custom should.js `equalInterval` assertion didn't work correctly w/ uncertain boundaries. I've fixed that now.

As for using `normalizeInterval` inconsistently, I _tried_ to make it consistent. As far as I can tell, every operation in `src/datatypes/interval.ts` that returns an `Interval` normalizes it first (e.g., `union`, `intersect`, `except`). We don't do it on `copy()` or `toClosed()`, but those functions are only used internally to support operations; they don't correspond directly to any operations themselves. We don't do it in the `expand` or `collapse` operations defined in `src/elm/interval.ts`, but I don't think we need to due to how they're implemented (keeping null boundaries intact).

But... as you noted, it's not strictly necessary. If we skip normalization, the intervals still represent the same range of values. So removing this does remove code complexity and ever-so-slightly improves performance. I think I'd be ok with removing it, but I'd like to know @lmd59's thoughts (or someone else from that team) about impact to MADiE and its users. Do you think this would have any impact there or affect usability? Since most measure outputs are about membership (or encounters), I don't think Intervals w/ null boundaries are likely in the primary outputs, but are they visible in debugging/highlighting?

There is a middle ground: make it configurable. But I'm not sure if there's much value in that. If the point of removing it is to reduce complexity, then making it configurable kind of defeats that point.

cmoesel:
I don't know why the CI failed, but I re-ran it (without changing anything) and it passed. I thought that would update the UI here, but I'm not seeing it. Here are the updated results: https://github.com/cqframework/cql-execution/actions/runs/30477306092?pr=373

src/datatypes/interval.ts
lmd59: This is an interesting question. From an fqm-execution perspective, I don't think we have a strong inclination in either direction.
For MADiE, I would want to check in with the MADiE team to understand how/whether this might affect them one way or another. @JSRankins - would you be able to weigh in on this discussion and provide any opinions on whether interval normalization would affect MADiE and its users?

lmd59 (approved):
Regression for fqm-execution looks good to me!

src/datatypes/interval.ts
dehall: Based on further discussion, @cmoesel's explanation and some more cogitation on my part, I changed my mind and now say: let's keep this interval normalization. We still welcome everyone's thoughts but since normalization is closest to how the library works in production today and is meant for the benefit of users, I'd say it's safe to merge in now, and if we find out it's not necessary later we can remove it.

But @cmoesel as one last thing, would you mind expanding the comment around this function to more directly explain why it exists? ie, because the functions that construct a new interval use `start` and `end` which return the closed value for the boundary, intervals that started out looking "simple" such as `[null, 0]` can result in new intervals with specific values that are not obvious to all users, such as `[-2147483648, 0]`. `normalizeInterval` recognizes those situations and converts the interval back to an equivalent form that is more likely to be familiar to all users. Normalized and "non-normalized" intervals are intended to be equivalent in all use cases except for manual inspection of the raw fields of the interval. (There are probably more precise terms for some of this)

cmoesel: @dehall - See [dd09d09](https://github.com/cqframework/cql-execution/pull/373/commits/dd09d09534cc9a50adf088892ec1eb41c99e4d2e). Does that address your concerns?

dehall: Yes, thank you!

PR #374, opened 2026-07-31
Updates to Dependencies, Build Chain, and Examples

cmoesel:
This PR makes non-trivial updates to some of our dependencies, our build chain, and our examples. These changes include:

* **Switch from ESLint to [Oxlint](https://oxc.rs/docs/guide/usage/linter.html).** ESLint has become more complicated and fragile over the years. Currently, it is keeping us from upgrading to TypeScript 7 and it's not clear if/when that will be unblocked (see [typescript-eslint#10940](https://github.com/typescript-eslint/typescript-eslint/issues/10940)). Oxlint is an extremely fast Rust-based linter built by the same team that built Vite. This change allowed us to replace _six_ ESLint-related packages with a _single_ OxLint package. I also removed a couple of TypeScript rules that I didn't think were necessary.
* **Upgrade TypeScript 6 to [TypeScript 7](https://devblogs.microsoft.com/typescript/announcing-typescript-7-0/).** With ESLint gone, we are also able to upgrade to TypeScript 7. TypeScript 7 is 10x faster than TypeScript 6 because it uses a brand new compiler written in Go.
* **Replace ts-node with [tsx](https://tsx.hirok.io/).** This was required by the switch to TypeScript 7, but since ts-node is no longer maintained, it makes sense to switch anyway.
* **Replace browserify with [esbuild](https://esbuild.github.io/).** Browserify was cool in 2015, but many web packing tools have surpassed it in performance and robustness. It was last updated in 2024 (and the last update before that was 2020). It has also caused npm audit issues in the past due to transitive dependencies.
* **Convert examples to full mini-projects.** We currently include three simple examples for different technologies: browser, node, and typescript. This PR makes it easier for users to build, run, and adapt them by providing them as small independent node projects.
* **Remove checked in cql4browsers.js.** Bonnie used to use cql4browsers.js directly, but nobody does that anymore. Now everyone uses their own web packing library (rollup, esbuild, etc) to integrate cql-execution into their web applications. Now the cql4browsers.js file is part of the build process for the browser example and is git-ignored. No more annoying "dang, I forgot to rebuild cql4browsers" moments!
* **Stop support for Node 18.** Somewhere along the line, Mocha stopped working with Node 18. The production cql-execution code technically still works with Node 18, but our build tools don't. Since Node 18 was EOLed more than a year ago, it probably makes sense just to drop it since we can't unit test with it anymore.

Some of these changes might be considered breaking (particularly dropping Node 18 and dropping cql4browsers), so we should consider if this warrants a major version change. In practice, I doubt this will break any existing implementations, but if we follow semver strictly, we probably _should_ bump it.

**Submitter:**

- [n/a] `cql4browsers.js` built with `npm run build:browserify` if source changed.

**Reviewer:**

Name:


examples/browser/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "cql-execution-browser-example",
+  "version": "1.0.0",
+  "private": true,
+  "type": "module",
+  "scripts": {
+    "build": "esbuild exec-age.js --bundle --platform=browser --format=iife --target=es2020 --outfile=dist/cql4browsers.js",
```
dehall: This is very minor but I would propose naming this output file something else. We're getting rid of the included cql4browsers so having anything refer to that name could potentially cause confusion.

dehall (approved):
This looks good to me and I'd be ok with merging as-is. But since it has been a couple weeks since this was submitted you may want to bump dependencies one more time just to get the latest.
My only concern is that this is the kind of PR where the things that go wrong are things I wasn't aware of, like how others interact with the library. But to your point, I think if we bump the major version number then anyone who is using this should be able to figure out any breaking changes.

examples/browser/package.json
cmoesel: Sure. That makes sense, especially since the built bundle has exec age logic in the entrypoint now. I renamed it to `dist/exec-age.js` in [74d23e5](https://github.com/cqframework/cql-execution/pull/374/commits/74d23e5f55bf3d15e43976759a92838fe00d2d2c).

cmoesel:
@dehall - your comments have been addressed. Based on your comment, I think you are advising we bump a major version, right? (BTW -- I'll also update the GH merge criteria to no longer check cql4browsers once this is ready to merge to master).

dehall:
> @dehall - your comments have been addressed. Based on your comment, I think you are advising we bump a major version, right? (BTW -- I'll also update the GH merge criteria to no longer check cql4browsers once this is ready to merge to master).

I guess to be precise I would say, I think that the next tagged release (whenever that is) should be a bump to the major version. I'm also thinking about my Decimal changes which will impact how users interact with this library, so I'd propose waiting for that before releasing the next version. The fqm-execution team also wants to do a more detailed regression test as well. (And let's also consider if there's anything else we want to try to get in before that. For example I'd kind of like to take a little time to do a quick pass of tightening up types and reducing the use of `any`)

hossenlopp:
Just chiming in here since I was able to run a large regression test on this PR vs master with fqm-execution. No changes with results or issues at all. This gets a thumbs up from me. 👍

hossenlopp (approved):

cmoesel:
Thank you for the reviews and testing, @dehall and @hossenlopp. Dylan -- holding off a release until the Decimal work (and possibly more breaking changes) sounds like a good plan.

